PR #3, opened 2022-06-14
10643-get-global-layers-before-the-event-geometry

IlyaIzr:


src/features/layers_in_area/atoms/areaLayersListResource.ts
```diff
@@ -67,18 +67,20 @@ const areaLayersListDependencyAtom = createAtom(
...
     eventId?: string;
     geoJSON?: GeoJSON.GeoJSON;
     eventFeed?: string;
     appId?: string;
-  } = {
-    geoJSON: params?.focusedGeometry.geometry,
-  };
+  } = params.focusedGeometry
+    ? {
+        geoJSON: params.focusedGeometry.geometry,
```
propakov, line 79: From top areaLayersListDependencyAtom it looks like  focusedGeometry  could be null. 
In that case `params.focusedGeometry.geometry,` will throw error
Maybe we need to process it on top, like it was, or return optional chaining

IlyaIzr: But it's checked in ternary operator just one line above:
if there's no `params.focusedGeometry`, empty object will be returned and we won't try to access `params.focusedGeometry.geometry`

propakov: ah, that's all right then!

propakov (approved):

PR #8, opened 2022-06-15
#10612 Extra graphql request while logging out from dn2

andrei-kucherov:
Fix. Remove s extra graphQL request while logging out from DN2 with the previously opened bivariate manager.


propakov (approved):
👍

PR #9, opened 2022-06-15
10620-focus-to-administrative-boundary-button-woks-incorrectly

IlyaIzr:
I moved part of our focusing to ~utils and added focusing to feature to boundary selector feature

### DIFF
--- a/src/features/boundary_selector/atoms/boundaryMarkerAtom.ts
+++ b/src/features/boundary_selector/atoms/boundaryMarkerAtom.ts
@@ -1,17 +1,17 @@
 import { createAtom } from '~utils/atoms';
 import { currentMapAtom, currentMapPositionAtom } from '~core/shared_state';
+import app_config from '~core/app_config';
 import { constructOptionsFromBoundaries } from '~utils/map/boundaries';
 import { convertToAppMarker } from '~utils/map/markers';
 import { sideControlsBarAtom, focusedGeometryAtom } from '~core/shared_state';
 import { i18n } from '~core/localization';
+import { getCameraForGeometry } from '~utils/map/cameraForGeometry';
 import { BOUNDARY_MARKER_ID } from '../constants';
 import { clickCoordinatesAtom } from './clickCoordinatesAtom';
 import { boundaryResourceAtom } from './boundaryResourceAtom';
 import { highlightedGeometry } from './highlightedGeometry';
 import { getSelectorWithOptions } from './../components/getSelectorWithOptions';
 import type { ApplicationMapMarker } from '~components/ConnectedMap/ConnectedMap';
-import { getCameraForGeometry } from '~utils/map/cameraForGeometry';
-import app_config from '~core/app_config';
 import type { Action } from '@reatom/core';
 
 const LOADING_OPTIONS = [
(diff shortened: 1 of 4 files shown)


src/features/boundary_selector/atoms/boundaryMarkerAtom.ts
```diff
@@ -10,6 +10,9 @@ import { boundaryResourceAtom } from './boundaryResourceAtom';
 import { highlightedGeometry } from './highlightedGeometry';
 import { getSelectorWithOptions } from './../components/getSelectorWithOptions';
 import type { ApplicationMapMarker } from '~components/ConnectedMap/ConnectedMap';
+import { getCameraForGeometry } from '~utils/map/cameraForGeometry';
```
propakov: doesn't eslint warn you in ide here?

andrei-kucherov (dismissed):

src/features/boundary_selector/atoms/boundaryMarkerAtom.ts
```diff
@@ -100,19 +97,38 @@ export const boundaryMarkerAtom = createAtom(
 
           const markerData = getSelectorWithOptions(
             options,
+
             // onOptionSelect:
             (boundaryId) => {
               if (!featureCollection) return;
+              const selectedFeature = featureCollection.features.find(
```
propakov, line 104: also it's possibly null here, is there any sense to process this case?

src/features/geometry_uploader/index.tsx
```diff
@@ -52,15 +51,9 @@ export function initFileUploader() {
...
-        const map = currentMapAtom.getState();
-        if (!map) return;
 
-        bbox.length = 4;
-        const camera = map.cameraForBounds(bbox, {
-          padding: app_config.autoFocus.desktopPaddings,
-        });
-        if (!camera) return;
-        const { zoom, center } = camera;
+        if (!geometryCamera) return;
```
propakov: i would expect to see this early return right before` if (typeof geometryCamera === 'string') {`

propakov (dismissed):
👍

src/features/boundary_selector/atoms/boundaryMarkerAtom.ts
IlyaIzr: `selectedFeature` can't be null here, featureCollection.features is array of features (they're objects of type GEOJson.Feature). Have you meant something else?

src/features/boundary_selector/atoms/boundaryMarkerAtom.ts
IlyaIzr: nope :(

src/features/geometry_uploader/index.tsx
IlyaIzr: fixed

src/features/boundary_selector/atoms/boundaryMarkerAtom.ts
propakov: I just ment that .find() can or cant find smth
if it can't -> selectedFeature will be null

src/features/boundary_selector/atoms/boundaryMarkerAtom.ts
IlyaIzr: resolving: fix delivered to main branch

propakov (approved):

src/features/boundary_selector/atoms/boundaryMarkerAtom.ts
IlyaIzr: I see, it could be `undefined` if nothing found, but there's `!` in the end of `.find()` method, meaning we're shure it will find something. I wrote that `!` because previous code implied that and because `boundaryId` and `featureCollection` that we are comparing against comes from the same source

PR #18, opened 2022-06-17
10443-layer-kontur-osm-quantity-was-highlighted-with-red-color

IlyaIzr:


vite.config.ts.js
```diff
@@ -0,0 +1,319 @@
+// vite.config.ts
```
Komzpa: Is it expected here?

andrei-kucherov (approved):

PR #21, opened 2022-06-22
fix(build): split dependencies and devDependencies

andrei-kucherov:


package.json
```diff
@@ -140,24 +124,36 @@
     "react-transition-group": "^4.4.2",
     "react-virtuoso": "^2.2.3",
     "rollup-plugin-visualizer": "^5.5.2",
-    "sinon": "^12.0.1",
-    "stylelint": "^14.1.0",
-    "stylelint-config-recommended": "^6.0.0",
-    "stylelint-config-standard": "^24.0.0",
     "typescript": "4.3.5",
```
propakov, line 128: "typescript": "4.3.5",

"postcss-custom-media": "^8.0.0",
"postcss-loader": "^6.1.1",
"postcss-nested": "^5.0.6",
"postcss-normalize": "8.0.1",

"patch-package": "^6.4.7",

@andrei-kucherov  what do you think, do we need these packages in dev dependencies? 
 it's usually expected to put transpilers to dev deps, and patch-package is called only when running dev stuff

src/features/side_bar/components/SideBar/SideBar.tsx
```diff
@@ -1,11 +1,11 @@
 import { useAction, useAtom } from '@reatom/react';
 import { ActionsBar, ActionsBarBTN } from '@konturio/ui-kit';
-import { v4 as uuidv4 } from 'uuid';
```
propakov, line 3: 👍

package.json
andrei-kucherov: This is was the question: should I move tools that needed to actually build the app into dev or into normal deps. If I will move these deps into dev deps should I move vite and all its plugins into dev deps as well?

propakov: I would expect to see builders in dev deps also, as they are not into final bundle.
https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file

so as they say:
"devDependencies": Packages that are only needed for local development and testing.

@Akiyamka @IlyaIzr guys, what's your opinion here?

Akiyamka: It looks to me that we should sync out understanding about what is dev dependencies is.
Foe me  "Packages that are only needed for local development and testing." it's just testing and linters tools, some debug/helper scripts. From that point of view vite and other tools that required for deployment i'ts not dev tools.
So in summary I see at leat more that 2 gourps of  deps:
- Deps that need only for development: vitest, sinon, eslint, stylint, husky, etc.
- Build time deps - vite, postcss, patch-package, etc.
- Runtime deps (will be included in bundle) - react, maplibre, nebula, etc

propakov (approved):

PR #22, opened 2022-06-23
10639 geometry changed it s shape after moving around the map

IlyaIzr:
Nebula GL used Translate Mode to move features on map
This mode used turfTransformTranslate(), giving it any geometry at all
And that caused serious deformations when moving polygons to poles

So I implemented a solution that moves polygons center first and then builds polygon around it. That way geometry distortion is significantly smaller and not really visible

### /CustomTransformMode.ts is a copy of nebula module with 1 import change
(to use local translation mode)


src/core/draw_tools/map-daw-tools/customDrawModes/translateFunction.test.ts
```diff
@@ -0,0 +1,152 @@
...
+  expect(result.geometry.coordinates[0][1]).toBeTypeOf('object');
+  expect(result.geometry.coordinates[0][1].length).toBeDefined();
+  expect(result.geometry.coordinates[0][2]).toBeTypeOf('object');
+  expect(result.geometry.coordinates[0][2].length).toBeDefined();
+
+  expect(result.geometry.coordinates[0][3]).toBeUndefined();
+  expect(result.geometry.coordinates[1]).toBeUndefined();
+});
+
+test('Polygon coordinates in right format', () => {
```
Akiyamka, line 114: Tests should have different names, How you understand from log what test is fail? Be more specific

src/core/draw_tools/map-daw-tools/customDrawModes/CustomModifyMode.ts
```diff
@@ -77,7 +78,7 @@ export class CustomModifyMode extends GeoJsonEditMode {
           this._submodesCache[submodeName] = new ModifyMode();
           break;
         case 'Transform':
-          this._submodesCache[submodeName] = new TransformMode();
+          this._submodesCache[submodeName] = new CustomTransformMode();
```
Akiyamka, line 81: Currently the only  way to understand difference between TransformMode and CustomTransformMode in line by line soruce code comparing. Need more details

Komzpa:
Can we fix this in Nebula upstream instead?

Akiyamka:
I think instead of making a custom mod we should make a fix in the nebula. After all, we want to get fixes from them too

IlyaIzr:
After preparing upstream MR to Nebula i found out there was possibly a way already to do the lossless translation. I'm trying to configure it now, as it's almost not documented and then compare to my solution

IlyaIzr:
Checked Nebula implemented ' without distortion ' translation. It just moves shape without any transormation according to projection
image

Komzpa:
This is not desired behavior as the shape is not becoming larger when going towards the pole.

IlyaIzr:
Pull request to Nebula: https://github.com/uber/nebula.gl/pull/772

Komzpa:
@IlyaIzr is there a screenshot/gif on what happens when you move geometry to far north? (it has to grow larger)

PR #24, opened 2022-06-24
hints implemented for bivariate legend

propakov:
https://kontur.fibery.io/Tasks/Task/Add-sentiment-hints-for-bivariate-legend-fe-10671

### DIFF
--- a/src/components/BivariateLegend/CornerTooltipWrapper.tsx
+++ b/src/components/BivariateLegend/CornerTooltipWrapper.tsx
@@ -11,7 +11,7 @@ import type { CornerRange } from '~utils/bivariate';
 
 export type CornerTooltipWrapperProps = {
   meta: BivariateLegendProps['meta'];
-  children: ReactNode | ReactNode[];
+  children: ReactNode;
 };
 
 const CORNER_POINTS_INDEXES = [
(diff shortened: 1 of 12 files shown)


src/utils/bivariate/bivariateLegendUtils.ts
```diff
@@ -103,5 +95,15 @@ export function createBivariateMeta(
   return {
     description,
     copyrights,
+    hints: {
+      x: {
+        label: xNominatorLabel,
+        directions: xAxisNominatorIndicator?.direction,
```
IlyaIzr: wierd that it's in plural - 'directions'

src/components/BivariateLegend/CornerTooltipWrapper.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { BIVARIATE_LEGEND_SIZE } from './const';
+import s from './CornerTooltipWrapper.module.css';
+import type { ReactNode } from 'react';
+import type { Cell } from '@konturio/ui-kit/tslib/Legend/types';
+import type { BivariateLegendProps } from './BivariateLegend';
+import type { CornerRange } from '~utils/bivariate';
+
+export type CornerTooltipWrapperProps = {
+  meta: BivariateLegendProps['meta'];
+  children: ReactNode | ReactNode[];
```
andrei-kucherov: Guess that ReactNode works for multiple children as well. Seems like ReactNode[] is redundant

propakov: hm, I just did the same way like we did in other places

andrei-kucherov (approved):

PR #26, opened 2022-06-27
9035 switch fe from user profile endpoint get features to dn be endpoint

andrei-kucherov:


propakov (approved):
👍

src/utils/atoms/createResourceAtom.ts
```diff
@@ -245,6 +245,8 @@ export function createResourceAtom<P, T>(
     }
   } else {
     resourceFetcherAtom = createResourceFetcherAtom<P, T>(fetcher, name);
+    // auto start resource atom that has no params atom controlling it's flow
+    resourceFetcherAtom.request.dispatch(null);
```
Akiyamka, line 249: Looks reasonable, but it's breaking change, i hope it's not broken something

Akiyamka (approved):

src/utils/atoms/createResourceAtom.ts
andrei-kucherov: Checked the code and found that we never used resource atoms without parameter atom before. So, hope that these changes will work smoothly

PR #31, opened 2022-06-28
10944 Check if configs have needed values

Akiyamka:
BREAKING CHANGE:
configs now in .json format
due to #10821 task
image


propakov (approved):
👍

PR #32, opened 2022-06-29
10453 tooltips opened in wrong places when change browser window size

IlyaIzr:


src/features/tooltip/Tooltip/Tooltip.tsx
```diff
@@ -1,38 +1,51 @@
...
-        else setPosition('top-right');
-      }
-      // click was on the top right side
-      else if (properties.position.x > width / 2) {
-        setPosition('bottom-left');
-      } else setPosition('bottom-right');
-    } else setPosition(null);
-  }, [properties?.position]);
+  if (
+    prevCoords?.x !== properties?.position.x &&
```
propakov: we have `properties: TooltipData | null; ` 
TooltipData.position is required field. So we are sure, that we have it always. 
at line 61 we process case  `if (!properties) return null;`
I suggest to do it at start, right after `useState` line 35, and get rid of all these optional chaining checks: properties?.position.

propakov (approved):
good job! 👍

src/features/tooltip/Tooltip/Tooltip.tsx
IlyaIzr: thanks, fixed it, I don't like unnescessary .? either

PR #39, opened 2022-07-02
11031-event-from-url-doesn-t-get-autofocus

IlyaIzr:


src/core/shared_state/currentEventFeed.ts
```diff
@@ -34,7 +34,7 @@ export const currentEventFeedAtom = createAtom(
     onChange('userResourceAtom', ({ data, loading, error }) => {
       if (!loading && !error && data && data.feeds && data.feeds.length) {
         const newFeed = data.checkFeed(state?.id);
-        if (newFeed !== undefined) {
+        if (newFeed !== undefined && newFeed !== state?.id) {
```
Akiyamka, line 37: It's a good check, but could you describe briefly how its affect autofocus?

Akiyamka (approved):

src/core/shared_state/currentEventFeed.ts
IlyaIzr: First of all I did bad task naming - it affects autofocus for event list panel (it's broken atm)
Why it happens? 
`currentEventAtom` handles `onChange`of `currentEventFeedAtom` incorrectly: at the start of the app we still don't have focus geometry that sets event to null. Thats okay. Then here in `currentEventFeed` feed changes from `{ id: 'default-id' }` to `{ id: 'default-id' }`, that triggers onChange handler, while focus geometry is still null, (but event was applied from url)
and resets current event to null. And that's noticable in url and event list focus stays and first default event

Akiyamka: got it

PR #48, opened 2022-07-05
fix bug: event id is removed from request for app with limited ammount of features

Akiyamka:
Reason of this bug was in autoselect logic (look at first commit in branch)
In scope of this fix i rewrite logic of autoselect and autozoom to more imperative style - look like it this case it more suitable


propakov (approved):

propakov:
👍

PR #52, opened 2022-07-07
Get data for list with sentiments combinations 

propakov:
https://kontur.fibery.io/Tasks/Task/Get-data-for-list-with-sentiments-combinations-10930

it's empty page on route /bivariate-manager
so we gonna put new layout here 

Screenshot 2022-07-07 at 12 09 14

### DIFF
--- a/src/features/bivariate_manager/atoms/bivariateStatisticsResource.ts
+++ b/src/features/bivariate_manager/atoms/bivariateStatisticsResource.ts
@@ -6,7 +6,6 @@ import {
   isGeometryEmpty,
 } from '~features/bivariate_manager/utils/createBivariateGraphQLQuery';
 import { parseGraphQLErrors } from '~features/bivariate_manager/utils/parseGraphQLErrors';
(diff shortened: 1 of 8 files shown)


src/features/bivariate_manager/atoms/bivariateStatisticsResource.ts
```diff
@@ -37,17 +39,17 @@ export const bivariateStatisticsResourceAtom = createResourceAtom(
           },
         );
       } catch (e) {
-        if (isApiError(e) && e.problem.kind === 'canceled') {
+        if (abortController.signal.aborted) {
```
andrei-kucherov: Hmm. Looks like that here we go from broad case (request can be cancelled by several cases, not just manual cancel in atom) to more specific case

andrei-kucherov (approved):

src/features/bivariate_manager/atoms/bivariateStatisticsResource.ts
propakov: fixed

PR #53, opened 2022-07-07
added icon with redirect to biv color manager page

propakov:
https://kontur.fibery.io/Tasks/Task/Add-icon-to-sidebar-to-access-color-management-section-dn-fe-10836

### DIFF
--- a/src/views/Main/Main.tsx
+++ b/src/views/Main/Main.tsx
@@ -92,6 +92,13 @@ export function MainView() {
         initReportsIcon(history),
       );
     }
+
+    if (userModel?.hasFeature(AppFeature.BIVARIATE_COLOR_MANAGER)) {
+      import('~features/bivariate_color_manager/').then(
+        ({ initBivariateColorManagerIcon }) =>
+          initBivariateColorManagerIcon(history),
+      );
+    }
     if (userModel?.hasFeature(AppFeature.OSM_EDIT_LINK)) {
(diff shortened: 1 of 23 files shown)


src/views/Main/Main.tsx
```diff
@@ -92,6 +92,13 @@ export function MainView() {
         initReportsIcon(history),
       );
     }
+
+    if (userModel?.hasFeature(AppFeature.BIVARIATE_COLOR_MANAGER)) {
+      import('~features/bivariate_color_manager/').then(
```
andrei-kucherov: Just thinking about the reason of using lazy import here. Now the logic is: we load **index** bundle, it checks features and if color management feature is available then it loads another bundle that consists of one file **src/features/bivariate_color_manager/index.tsx** after this file runs its logic, app redirects user to new route **./bivariate-manager** and after that our router will load another bundle with actual bivariate color config (we also use lazy loading in router, so another bundle will be created consequently). The question here is: do we really need to split **bivariate_color_manager/index.tsx** into a separate bundle?

andrei-kucherov (approved):

src/views/Main/Main.tsx
Akiyamka: good point

propakov: You're right! This logic looks redundant. 
I also see we have the same situation with  import('~features/reports/').
So I'll remove lazy-loading for it also.

andrei-kucherov (approved):

PR #54, opened 2022-07-08
config fix for report table page

IlyaIzr:


src/core/app_config/loader.ts
```diff
@@ -1,5 +1,9 @@
 export async function loadConfig() {
-  const response = await fetch('config/appconfig.json');
+  const pathname = window.location.pathname;
```
andrei-kucherov: Guess that `const response = await fetch(`${import.meta.env?.BASE_URL}config/appconfig.json`);` can be a bit more efficient here (Will cover all cases, not just /active/).

propakov (approved):

PR #55, opened 2022-07-08
10490 empty reports should not break frontend

IlyaIzr:


propakov (approved):
👍

src/features/reports/components/ReportTable/ReportTable.tsx
```diff
@@ -23,15 +26,24 @@ function openOSMID(
...
+      setState({ sortIndex: 0, ascending: null });
     };
   }, []);
 
+  if (data === null) {
+    return <Text type="heading-xl">{i18n.t('No data for this report')}</Text>;
+  }
+
+  if (data === 'sorting') {
+    return <Text type="heading-xl">{i18n.t('Sorting data...')}</Text>;
```
propakov, line 44: should we put the translations in src/core/localization/translations/en/common-messages.json with the key?
or it's ok to go this way?

IlyaIzr: maybe we should but that's another discussion, we've mostly used this simple way that far. I've added this point to discuss on next fronted meetup

PR #63, opened 2022-07-15
Research bug with map blinking when we get default id 11126

Akiyamka:


IlyaIzr (approved):

src/Routes.tsx
```diff
@@ -32,41 +32,44 @@ const ROUTES = {
 };
 
 export function RoutedApp() {
-  const [{ data: userModel }] = useAtom(userResourceAtom);
+  useEffect(() => {
```
propakov, line 35: we need route params only on map page
this way it will add them on every route after refresh

I suggest to return initUrlStore like it was, but in this file we already have 
`currentApplicationAtom.init.dispatch();`
https://github.com/konturio/disaster-ninja-fe/pull/63/files#diff-504a28a4c73adc1fe14d06359099ab7a49405f8705eb8c00214cbd5a94bfde80R116

we just need to let it be called so let's do condition this way

`<Router><CommonRoutesFeatures userModel={userModel} />
          {userModel && !loading && (
          <CacheSwitch>`

PR #66, opened 2022-07-15
11062 add use map view as filter control in disasters panel

Akiyamka:
feat(events_list)!: 11062, 11060, 10093, 11058
add filtering by bbox of current view

BRAKING CHANGE:
currentMapAtom empty state now `null` instead of `undefined`


src/features/events_list/components/FeedSelector/FeedSelector.tsx
```diff
@@ -12,17 +13,20 @@ const FeedSelectorComp = () => {
...
+    [setCurrentFeed, scheduleAutoSelect],
+  );
 
   return userModel &&
     userModel.hasFeature(AppFeature.FEED_SELECTOR) &&
     userModel.feeds &&
     userModel.feeds.length > 1 ? (
     <div className={s.feedSelectorContainer}>
-      <div>{i18n.t('Feed')}:</div>
+      <Text type="short-m" /* type="label" */>{i18n.t('Feed')}:</Text>
```
propakov: do we need this comment? /* type="label" */

src/utils/atoms/combineAtoms.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function combineAtoms<T extends AtomsMap>(
+  shape: T,
+): Atom<{ [key in keyof T]: T[key] extends Atom<infer S> ? S : never }> {
+  return createAtom(shape, ({ get }, state = {}) => {
+    return Object.keys(shape).reduce((acc, key) => {
+      // @ts-ignore
+      acc[key] = get(key);
+      return acc;
+    }, {} as { [key in keyof T]: T[key] extends Atom<infer S> ? S : never });
```
propakov: isn't it unknown instead of never here ?
if so, it's possible to go this way, to eliminate // @ts-ignore
```

export function combineAtoms<T extends AtomsMap>(
  shape: T,
): Atom<{ [key in keyof T]: T[key] extends Atom<infer S> ? S : unknown }> {
  return createAtom(shape, ({ get }, state = {}) => {
    return Object.keys(shape).reduce((acc, key) => {
      acc[key as keyof T] = get(key);
      return acc;
    }, {} as { [key in keyof T]: T[key] extends Atom<infer S> ? S : unknown });
  });
}

```

propakov (approved):
👍 
pls consider one question about types, maybe it helps

src/features/events_list/atoms/eventListResource.ts
```diff
@@ -1,18 +1,37 @@
-import { createResourceAtom } from '~utils/atoms';
+import { createResourceAtom, combineAtoms } from '~utils/atoms';
 import { apiClient } from '~core/apiClientInstance';
 import { autoRefreshService } from '~core/autoRefreshServiceInstance';
 import { currentEventFeedAtom } from '~core/shared_state';
 import type { Event } from '~core/types';
+import { eventListFilters } from './eventListFilters';
```
propakov, line 5: it looks like eslint fix not working 🤷‍♂️

src/features/events_list/atoms/eventListResource.ts
```diff
@@ -1,18 +1,37 @@
...
+  filters: eventListFilters,
+});
 
 export const eventListResourceAtom = createResourceAtom(
-  async (currentFeed) => {
-    const params =
-      currentFeed && currentFeed.id ? { feed: currentFeed.id } : undefined;
-    const responseData = await apiClient.get<Event[]>(`/events/`, params, true);
+  async (depsAtomState) => {
+    const deps = depsAtomState!;
```
propakov: doesn't eslint alerts you here? 
Screenshot 2022-07-19 at 12 35 40

propakov: we can do 
```

    const deps = depsAtomState;

    const params: {
      feed?: string;
      bbox?: string;
    } = {
      feed: deps?.currentFeed?.id,
      bbox: deps?.filters.bbox?.join(','),
    };
```

src/features/events_list/components/FeedSelector/FeedSelector.tsx
Akiyamka: It should be label, but our kit not support it yet.
But i agree, in extra, i will remove it

propakov: I'm ok to stay it, no problem really, just was interested of the reason :)

src/utils/atoms/combineAtoms.ts
Akiyamka: Nice catch!

src/features/events_list/atoms/eventListResource.ts
Akiyamka: hmm, it's more look like lint-staged problem.
After I run `npx eslint --fix ./src` manually it was fixed

src/features/events_list/atoms/eventListResource.ts
Akiyamka: > doesn't eslint alerts you here? 

It is. Right now i'am sure that deps atom state can't be null, but typescript can't see that. I decide to avoid extra runtime checking in that case

propakov: ok, got it 👌

Akiyamka: Hm, probably i wrong, will use checks

PR #69, opened 2022-07-19
10805 event preview image generator add event to dn when page is ready

Akiyamka:


src/core/metrics/app-metrics.ts
```diff
@@ -3,25 +3,52 @@ import { Sequence } from './sequence';
...
+  private completed: Record<string, number> = {};
+  private eventLog: string[] = [];
+  private settings = new SessionSettings<'KONTUR_SQ_ALERT' | 'KONTUR_SQ_LOG'>();
 
   constructor() {
     globalThis.KONTUR_METRICS = {
       sequences: this.sequences,
       events: this.eventLog,
+      toggleAlert: () => this.settings.toggle('KONTUR_SQ_ALERT'),
+      toggleLog: () => this.settings.toggle('KONTUR_SQ_LOG'),
```
IlyaIzr, line 37: Why theese needed for global window object?

IlyaIzr (approved):
Works as expected although the whole thing is hard to grasp

src/core/metrics/app-metrics.ts
Akiyamka: Mostly for testers and debugging. Testers and back-ends want to see is event work right on fe side, so I add such "debug tools". Now the can write just `KONTUR_METRICS.toggleAlert()` for test event, and `toggleLog` to collect logs for task

Akiyamka:
@IlyaIzr Can you tell me which parts were difficult? I would revise them or at least add a commentary

IlyaIzr:
> @IlyaIzr Can you tell me which parts were difficult? I would revise them or at least add a commentary

Makes more sense to me now when I read it second time. in `src/core/metrics/sequences` set of string events looks somewhat random, may be some naming structure there would've helped

Akiyamka:
> > @IlyaIzr Can you tell me which parts were difficult? I would revise them or at least add a commentary
> 
> Makes more sense to me now when I read it second time. in `src/core/metrics/sequences` set of string events looks somewhat random, may be some naming structure there would've helped

agree, in can be improved by some atom naming convention -  right now they have pretty random names.
Inside reatom action name it string `${atomName}_${actionName}`

andrei-kucherov (approved):

PR #72, opened 2022-07-21
11062 add use map view as filter control in disasters panel patch

Akiyamka:
- Fix client typings (body null on 204 by fact vs undefined by types), and related errors
- Hide "Unknown error" stuff from user in event list panel

### DIFF
--- a/src/features/events_list/atoms/eventListResource.ts
+++ b/src/features/events_list/atoms/eventListResource.ts
@@ -20,8 +20,8 @@ export const eventListResourceAtom = createResourceAtom(
       bbox: deps?.filters.bbox?.join(','),
     };
 
-    const responseData = await apiClient.get<Event[]>('/events/', params, true);
-    if (responseData === undefined) throw new Error('No data received');
+    const responseData =
+      (await apiClient.get<Event[]>('/events/', params, true)) ?? [];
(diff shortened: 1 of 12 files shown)


src/features/events_list/atoms/eventListResource.ts
```diff
@@ -20,8 +20,8 @@ export const eventListResourceAtom = createResourceAtom(
       bbox: deps?.filters.bbox?.join(','),
     };
 
-    const responseData = await apiClient.get<Event[]>('/events/', params, true);
-    if (responseData === undefined) throw new Error('No data received');
+    const responseData =
+      (await apiClient.get<Event[]>('/events/', params, true)) ?? [];
```
propakov, line 24: maybe it's a good idea to process undefined also by `|| [];` ?

propakov (approved):

src/features/events_list/atoms/eventListResource.ts
Akiyamka: It depends on which way you look at it. In terms of defensive programming it is better to check for all invalid values including type (for example number 1 will pass the || check), and this how we usually do in pure js.
in terms of performance and predictable behavior it is better to keep types correct and be specific about what the code does, it's more typescript way

PR #74, opened 2022-07-22
feat(bivariate-color-manager): 10929 - implemented sublist

propakov:
https://kontur.fibery.io/Tasks/Task/Sublist-with-indicators-(Color-legends-section)-dn-fe-10929
Screenshot 2022-07-22 at 12 08 22
9

### DIFF
--- a/src/features/bivariate_color_manager/components/SentimentsCombinationsList/SentimentsCombinationsList.tsx
+++ b/src/features/bivariate_color_manager/components/SentimentsCombinationsList/SentimentsCombinationsList.tsx
@@ -1,13 +1,24 @@
-import { nanoid } from 'nanoid';
 import clsx from 'clsx';
-import { memo } from 'react';
+import React, {
+  Children,
+  cloneElement,
+  isValidElement,
+  memo,
+  useRef,
+  useState,
+} from 'react';
+import { CSSTransition } from 'react-transition-group';
 import { joinAndCapitalizeItems, sortByKey } from '~utils/common';
 import { i18n } from '~core/localization';
 import { MiniLegend } from '~features/bivariate_color_manager/components/MiniLegend/MiniLegend';
 import { invertClusters } from '~utils/bivariate';
 import s from './SentimentsCombinationsList.module.css';
-import type { BivariateColorManagerData } from '~features/bivariate_color_manager/atoms/bivariateColorManagerResource';
+import type {
+  BivariateColorManagerData,
+  TableDataValue,
+} from '~features/bivariate_color_manager/atoms/bivariateColorManagerResource';
 import type { BivariateLegend } from '~core/logical_layers/types/legends';
+import type { CSSTransitionProps } from 'react-transition-group/CSSTransition';
 
 type Row = {
   key: string;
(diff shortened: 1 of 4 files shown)


Akiyamka:
@propakov looks like link to task not correct

src/features/bivariate_color_manager/components/SentimentsCombinationsList/SentimentsCombinationsList.tsx
```diff
@@ -1,36 +1,57 @@
...
 const convertDirectionsArrayToLabel = (directions: string[][]) => {
   const [from = '', to = ''] = directions;
   return `${joinAndCapitalizeItems(from)} → ${joinAndCapitalizeItems(to)}`;
 };
 
+const fadeClassNames = {
+  enter: clsx(s.fadeEnter),
+  enterActive: clsx(s.fadeEnterActive),
+  exit: clsx(s.fadeExit),
+  exitActive: clsx(s.fadeExitActive),
```
Akiyamka: Record `enter: clsx(s.fadeEnter)` not make any sense, is the same as `enter: s.fadeEnter`
The reason why we use clsx is that you can write dynamic, conditional styles - like
```
className={clsx({ [s.fadeEnter]: isSomeBooleanProps, [s.fadeEnterActive]:  isSomeBooleanStateValue })}
```
also in easy way for join multiple classes: `clsx(s.classA, s.classB)` instead of `[s.classA, s.classB].join(' ')`

src/features/bivariate_color_manager/components/SentimentsCombinationsList/SentimentsCombinationsList.tsx
```diff
@@ -70,28 +91,91 @@ const SentimentsCombinationsList = memo(
...
 );
 
+const CSSTransitionWrapper = ({ children, ...props }: CSSTransitionProps) => {
+  const nodeRef = useRef(null);
+
+  return (
+    <CSSTransition {...props} nodeRef={nodeRef}>
+      <>
+        {Children.map(children, (child) =>
+          isValidElement(child) ? cloneElement(child, { ref: nodeRef }) : child,
```
Akiyamka, line 172: Be careful with `cloneElement`, is expensive operation, but here it clone only one children - tr, so it should be okay

Akiyamka (approved):

src/features/bivariate_color_manager/components/SentimentsCombinationsList/SentimentsCombinationsList.tsx
propakov: oh, sure
thanks for this catch!

src/features/bivariate_color_manager/components/SentimentsCombinationsList/SentimentsCombinationsList.tsx
propakov: do you know a better way of passing some prop to children? 
maybe render function pattern would work better?

```
<CSSTransitionWrapper
                  in={rowSelected}
                  timeout={300}
                  unmountOnExit
                  appear
                  classNames={fadeClassNames}
                >
                (ref) => 
                  <tr ref={ref} className={s.sublist}>
```

propakov:
> @propakov looks like link to task not correct

sorry, last digit has lost
I've updated the link!

PR #78, opened 2022-07-28
fix(core): 11085 - layers weren't visible on the map after authorization

IlyaIzr:


src/core/auth/atoms/userResource.ts
```diff
@@ -25,16 +25,23 @@ const userResourceRequestParamsAtom = createAtom(
...
     const applicationId = get('currentApplicationAtom');
     const userData = get('currentUserAtom');
 
-    return {
+    // if user profile had changed
+    if (state && state.userData?.username !== userData.username) {
+      // mark all layers as unmounted
+      schedule((dispatch) => dispatch(mountedLayersAtom.clear()));
+    }
+    state = {
```
Akiyamka: it would make sense if those layers were part of the user's information, but they are not.
We need to fix the real reason

Akiyamka (changes requested):
Fix the reason

src/core/auth/atoms/userResource.ts
IlyaIzr: I don't quite understand. What layers should be part of user information? 
When user changes (logs in or out) most of our app redraws, that includes map - so all layers become unmounted. Here I am marking them as unmounted.

Akiyamka: > What layers should be part of user information?  

They are not, and the code says otherwise

Akiyamka (approved):

PR #80, opened 2022-08-01
feat(bivariate-legend): 10909 add hints for presents for bivariate

propakov:
https://kontur.fibery.io/Tasks/Task/Add-sentiment-hints-for-bivariate-legend-for-presets-fe-10909

### DIFF
--- a/src/utils/bivariate/types/stat.types.ts
+++ b/src/utils/bivariate/types/stat.types.ts
@@ -18,11 +18,17 @@ export type Step = {
 };
 
 /* Divisor and denominator pair */
-export type Quotient = [string, string];
+export type Quotient = [string, string]; // this field will be removed in next tickets, only Quotients will stay
+export type Quotients = {
+  name: string;
+  label: string;
+  direction: Direction;
+};
 
 export type Axis = {
   steps: Step[];
   quotient: Quotient;
+  quotients?: Quotients[];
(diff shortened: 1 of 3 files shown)


src/utils/bivariate/types/stat.types.ts
```diff
@@ -19,10 +19,16 @@ export type Step = {
 
 /* Divisor and denominator pair */
 export type Quotient = [string, string];
+export type Quotients = {
```
IlyaIzr: Very similar names. What both of them means? I think their names should give move info or some comments should be placed around

IlyaIzr (changes requested):
other than naming I approve

src/utils/bivariate/types/stat.types.ts
propakov: hm, I see. 
This is because it's added in response on backend side
https://kontur.fibery.io/Tasks/Task/Add-info-about-indicator-sentiments-and-labels-for-presets-on-DN2-be-10722
So now we have quotient  and quotients. Types describes what's inside, so maybe put a comment here?
Or maybe change name to QuotientDetails ?

IlyaIzr: One of those, don't know which one's better

IlyaIzr (approved):

PR #88, opened 2022-08-04
fix(core): 11438 - update url after going back to main from other pages via routing

IlyaIzr:


src/Routes.tsx
```diff
@@ -116,7 +116,7 @@ const CommonRoutesFeatures = ({ userModel }: CommonRoutesFeaturesProps) => {
       currentApplicationAtom.init.dispatch();
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [pathname]);
```
Akiyamka, line 119: @propakov why you ignore eslint here?

propakov: @Akiyamka Because I wanted to call it only once the app loaded.
To have hook like coomponentDidMount we need [] as dependencies, and there is no way without // eslint-disable-next-line  to make it pass

propakov: from my experience it's ok to skip react-hooks/exhaustive-deps if you really need it

propakov (approved):
@IlyaIzr pls be sure that urlStoreAtom.subscribe(() => null); called multiple times on route change won't affect app. Other than that looks good.

IlyaIzr:
> @IlyaIzr pls be sure that urlStoreAtom.subscribe(() => null); called multiple times on route change won't affect app. Other than that looks good.

fixed that and now merging

PR #92, opened 2022-08-09
feat(bivariate-color-manager): 10840 - filter by layers

propakov:
https://kontur.fibery.io/Tasks/Task/Filter-by-layers-(Color-legends-section)-dn-fe-10839
Screenshot 2022-08-09 at 16 29 55

### DIFF
--- a/src/features/bivariate_color_manager/atoms/bivariateColorManager.ts
+++ b/src/features/bivariate_color_manager/atoms/bivariateColorManager.ts
@@ -2,11 +2,27 @@ import { createAtom } from '~utils/atoms';
 import { bivariateColorManagerResourceAtom } from './bivariateColorManagerResource';
 import type { TableDataValue } from './bivariateColorManagerResource';
 import type { BivariateColorManagerData } from './bivariateColorManagerResource';
+import type { Indicator } from '~utils/bivariate';
+import type { SelectItemType } from '@konturio/ui-kit/tslib/Select/types';
+import type { BivariateColorManagerDataValue } from './bivariateColorManagerResource';
 
 export type BivariateColorManagerAtomState = {
-  data?: BivariateColorManagerData | null;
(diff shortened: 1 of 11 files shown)


src/features/bivariate_color_manager/atoms/bivariateColorManager.ts
```diff
@@ -16,25 +32,128 @@ export type LayerSelectionInput = {
...
+        // check if row passes all filter functions
+        const filterPassed = filterFunctionsToApply.every(
+          ([filterFunction, filterValue]) =>
+            filterFunction(key, value, filterValue),
+        );
+
+        if (filterPassed) {
+          // do postprocessing for passed row here
+
+          // if filter layer selected - stay only filtered layer in row's sublist
```
andrei-kucherov, line 120: A little bit confused that specific layer filtering logic take about a half of generic applyFilters method while there is already dedicated function **layersFilterFunction** for such purposes

andrei-kucherov (approved):

src/features/bivariate_color_manager/atoms/bivariateColorManager.ts
propakov: @andrei-kucherov the reason I've splitted it into 2 parts is next: 
- filterFunctionsToApply as an array of generic filter functions that will we applied to data array to filter it (it's pure)
- but we also need to modify it, it's not just filtering, so I decided to separate it to post filtering phase (not pure)
just don't want to mix pure and not pure staff, as it can lead to bugs

so as I see this: 
you need to add new filter - just add filter logic to filterFunctions and add filter body to the atom's store
if you need to modify smth after filtering - add postprocessing logic when specific filter applied

also as a case it's possible to implement this filtering sublist on component level
so we can get rid of this postprocessing part here 
but not sure it's a best approach

PR #93, opened 2022-08-10
feat(bivariate-color-manager): 11535 - fix sorting of sublist items.

propakov:
Now we collect avgCorrelationY by yQuotientIndicator.name and use it for sorting sublist items. Before we just used avgCorrelationY for both X and Y sublist items.

before
image

now after fix
image

### DIFF
--- a/src/features/bivariate_color_manager/atoms/bivariateColorManagerResource.ts
+++ b/src/features/bivariate_color_manager/atoms/bivariateColorManagerResource.ts
@@ -13,7 +13,7 @@ import type { LayerMeta } from '~core/logical_layers/types/meta';
 export type TableDataValue = {
   label: string;
   name: string;
-  quality?: number;
+  correlationLevel?: number;
   mostQualityDenominator?: string;
 };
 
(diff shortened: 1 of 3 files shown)


src/features/bivariate_color_manager/atoms/bivariateColorManagerResource.ts
```diff
@@ -227,3 +240,16 @@ export const bivariateColorManagerResourceAtom = createResourceAtom(
...
+  numeratorCorellationMap: NumeratorCorellationMap,
+): void => {
+  Object.values(bivariateColorManagerData).forEach((row) => {
+    [...Object.values(row.horizontal), ...Object.values(row.vertical)].forEach(
+      (layer) => {
+        layer.correlationLevel = numeratorCorellationMap[layer.name] || 0;
+      },
+    );
+  });
+};
```
Akiyamka: Just note:
Pretty unusual for me when function declared after they was used

Akiyamka (approved):

src/features/bivariate_color_manager/atoms/bivariateColorManagerResource.ts
propakov: this approach is from https://marekhudyma.com/code-style/2021/03/02/step-down-rule.html

so when you open file, key and high level content is on top, some util functions that usually about details(low level) are below

propakov: or sometimes I just unintentionally follow this rule
I'll put it on the top

Akiyamka: pretty interesting rule

PR #94, opened 2022-08-10
feat(ci): 11376 replace nginx with go server in front image 

Akiyamka:
also contain small fix some small issues in vite config
image


kalenikaliaksandr (approved):

index.html
```diff
@@ -19,7 +19,7 @@
   <meta name="msapplication-TileColor" content="#da532c">
   <meta name="msapplication-config" content="/favicon/browserconfig.xml?v=2.0.22">
   <meta name="theme-color" content="#ffffff">
-  <meta property="og:image" content="/active/preview.png?{{currentQueries}}">
+  <meta property="og:image" content="/active/preview.png?{{.CurrentQueries}}">
```
propakov, line 22: I see that main idea is to pass CurrentQueries here.
but can't find where is this file located? (active/preview.png)

vite.config.ts
```diff
@@ -11,19 +11,27 @@ import { buildScheme, validateConfig } from './scripts/build-config-scheme.mjs';
...
-import packageJson from './package.json';
 
 const relative = (folder: string) => path.resolve(__dirname, folder);
+const parseEnv = (
+  env: Record<string, string>,
+): Record<string, string> =>
+  Object.entries(env).reduce((acc, [k, v]) => {
+    try {
+      acc[k] = JSON.parse(v);
+    } catch (e) {}
```
propakov, line 22: why to have `catch` if nothing handled?

propakov (approved):
👍

index.html
Akiyamka: it address of image generator backend

vite.config.ts
Akiyamka: Because if it just string - i want to leave it as is.
```
JSON.parse('true')  // true
JSON.parse('false') // false
JSON.parse('null')  // null
JSON.parse('1')     // 1
JSON.parse('foo')   // throw Error
```

propakov: I ment to add console.log in catch block @Akiyamka

PR #98, opened 2022-08-11
add(layers_panel): 11563 - control exceptions for layers panel

IlyaIzr:
Visibility controls are now exist within Legend Panel only and removed from Layers Panel accorging to task 11563
image

### DIFF
--- a/src/features/layers_panel/components/Layer/useControlElements.tsx
+++ b/src/features/layers_panel/components/Layer/useControlElements.tsx
@@ -13,12 +13,18 @@ export function useControlElements(
     show: () => void;
     download: () => void;
   },
+  skipControls?: {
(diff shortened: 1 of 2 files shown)


src/features/layers_panel/components/Layer/useControlElements.tsx
```diff
@@ -13,12 +13,18 @@ export function useControlElements(
     show: () => void;
     download: () => void;
   },
+  skipControls?: {
```
Komzpa, line 16: I am more used to positive APIs than to negative ones, if that's just me you can dismiss this.

propakov (approved):

Komzpa (approved):
Design ok

Issue #101, opened 2022-08-12
Layers mount on map in unpredictable manner


IlyaIzr:
The issue is: some layers sometimes doesn't load on map. See this test stage case I've caught
![image](https://user-images.githubusercontent.com/50058726/184335396-baf6bcde-542e-4ca9-a1ad-7ffde3d8a2db.png)
This error happends 'cause we were trying to add source to map that wasn't done loading.
We've fixed that in `willMount` lifecycle hook for both generic and bivariate layers. But here it was caused by `willSourceUpdate` hook.

So the natural solution for me was to move fix one step higher, something like that

```
  private async _updateMap(
    map: ApplicationMap,
    layerData: LayerSource,
    legend: LayerLegend | null,
    isVisible: boolean,
  ) {
    if (layerData == null) return;
// adding this chunk
    if (!map.loaded()) {
      (await waitMapEvent(map, 'load'));
    }
// the wait here can become infinite
    if (this.isGeoJSONLayer(layerData)) {
      this.mountGeoJSONLayer(map, layerData, legend);
    } else {
      this.mountTileLayer(map, layerData, legend);
    }
```

But that wouldn't work - map won't mount layers. Now the interesting part - let's log our `waitMapEvent` helper

```
export const waitMapEvent = <T extends keyof MapEventType>(
  map: ApplicationMap,
  event: T,
) =>
  new Promise<MapEventType[T] & EventData | void>((res, rej) => {
    try {
      console.log('%c⧭ is map loaded already?', 'color: #735656', map._loaded, map.loaded());
      map.on(event, res);
    } catch (error) {
      rej(error);
    }
  });

```
When mounting fails following console logs can be seen
![image](https://user-images.githubusercontent.com/50058726/184341127-c31c8259-e3bd-4619-b821-2124eb384dd2.png)

This matters **hugely**, because `map.on('load', callback)` will not run, if it was setted after the map was actually loaded. 
And if we need to rely on map readyness (and we do) we need to rely on `map._loaded` property 

In other words example above could be working with on line fix 
```
  private async _updateMap(
    map: ApplicationMap,
    layerData: LayerSource,
    legend: LayerLegend | null,
    isVisible: boolean,
  ) {
    if (layerData == null) return;
// adding this chunk
    if (!map._loaded) {
      (await waitMapEvent(map, 'load'));
    }
// the wait here can become infinite
    if (this.isGeoJSONLayer(layerData)) {
[...]

PR #102, opened 2022-08-12
fix(core - logical_layers): 11657 - wait for map readiness in more reliable manner

IlyaIzr:
Explanation of why i'm using such approach: https://github.com/konturio/disaster-ninja-fe/issues/101

Problem we had:
image
Now works as expected

### DIFF
--- a/src/utils/map/waitMapEvent.ts
+++ b/src/utils/map/waitMapEvent.ts
@@ -13,7 +13,7 @@ export const waitMapEvent = <T extends keyof MapEventType>(
     }
   });
 
-export async function mapLoaded(map: ApplicationMap) {
+export function mapLoaded(map: ApplicationMap) {
   // Use hidden maplibre method for reliability. Explanation - https://github.com/konturio/disaster-ninja-fe/issues/101
   // @ts-expect-error
   if (!map._loaded) {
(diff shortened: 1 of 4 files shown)


andrei-kucherov (approved):

src/utils/map/waitMapEvent.ts
```diff
@@ -5,10 +5,19 @@ export const waitMapEvent = <T extends keyof MapEventType>(
...
-  new Promise<MapEventType[T] & EventData>((res, rej) => {
+  new Promise<(MapEventType[T] & EventData) | void>((res, rej) => {
     try {
       map.on(event, res);
     } catch (error) {
       rej(error);
     }
   });
+
+export async function mapLoaded(map: ApplicationMap) {
```
propakov: do we need async word if we don't use await inside?

propakov (approved):
good approach! 👍

Akiyamka:
Since we already fix layers mounting problem few times it looks to me that Generic render nice candidate for covering by tests

IlyaIzr:
> Generic render nice candidate for covering by tests

I completly agree considering this class also asks for refactoring

IlyaIzr:
follow up tasks were created - cover generic renderer with tests - 11725 and investigate difference between map._loaded and map.loaded() - 11726

PR #105, opened 2022-08-13
fix(draw_tools): 11045 - fix draw modes buttons behaviour

IlyaIzr:


src/core/draw_tools/components/DrawToolsToolbox/DrawToolsToolbox.tsx
```diff
@@ -38,29 +38,45 @@ export const DrawToolsToolbox = () => {
...
 
   const buttons = useMemo(
     () => [
       settings.availableModes?.includes('DrawPolygonMode') && (
-        <Button id={drawModes.DrawPolygonMode} dark variant="invert">
+        <Button
+          id={drawModes.DrawPolygonMode}
+          dark
+          variant="invert"
+          onClick={() => onToggleDrawMode(drawModes.DrawPolygonMode)}
```
propakov: you can just use toggleDrawMode as it will be new callback every render 

`const onToggleDrawMode = useCallback(`
this one can be removed at all

propakov (approved):
pls consider one comment

PR #106, opened 2022-08-13
add(header): 11568 - add about link to header

IlyaIzr:
image


Komzpa:
I don't see "chat with us" on the screenshot - how does it look together with the link? (maybe we need to match its font)

IlyaIzr:
Updated styles in order to match "Chat with us" button better
image

IlyaIzr:
Maybe it should be white with blue on hover? 
image
image

Komzpa (approved):

PR #109, opened 2022-08-16
fix(layers): 10798 - fix when two layers stuck on legend while fast swithing on slow internet

IlyaIzr:
cancel was called simply every time, i've changed that - it compares version and findes out if request was realy canceled

it fixes this
image

### DIFF
--- a/src/utils/atoms/createResourceAtom.ts
+++ b/src/utils/atoms/createResourceAtom.ts
@@ -92,7 +92,6 @@ function createResourceFetcherAtom<P, T>(
         newState.canceled = false;
         schedule(async (dispatch, ctx: ResourceCtx<P>) => {
           const version = (ctx.version ?? 0) + 1;
-          const nextParams = ctx.lastParams || null;
           ctx._refetchable = true;
           ctx.version = version;
           ctx.lastParams = params; // explicit set that request no have any parameters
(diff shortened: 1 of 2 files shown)


src/utils/atoms/createResourceAtom.ts
```diff
@@ -135,6 +135,13 @@ function createResourceFetcherAtom<P, T>(
               requestAction = create('error', e.message ?? e ?? true);
             }
           } finally {
+            // that's when request was canceled
+            if (version !== ctx.version) {
+              // extra dispatch allows resource subscribers to proccess cancel event and then
+              // process the response of the next event
+              if (ctx.allowCancel)
+                dispatch(create('cancel', nextParams, params));
```
andrei-kucherov: the logic is a bit complicated here and I may not be correct but it looks like the code here can be like:
`dispatch(create('cancel', ctx.lastParams, params));` 
where ctx.lastParams is the latest params of new request and params are previous request params

src/utils/atoms/createResourceAtom.ts
```diff
@@ -90,6 +92,7 @@ function createResourceFetcherAtom<P, T>(
         newState.canceled = false;
         schedule(async (dispatch, ctx: ResourceCtx<P>) => {
           const version = (ctx.version ?? 0) + 1;
+          const nextParams = ctx.lastParams || null;
```
andrei-kucherov: see my comment for line 143. probably this line not needed. you assign **nextParams** to **lastParams** while regular **params** have more recent values and surely this parameter can be even more outdated after async fetch. so, by the time this parameter is needed (line 143) **ctx.lastParams** will probably have newer value so better use it directly

src/utils/atoms/createResourceAtom.ts
IlyaIzr: `lastParams` confuses me a bit
what if here we would call them `nextParams` and `prevParams` so that we would definetly see the difference? Or `latestsParams` and `prevParams`?

andrei-kucherov (approved):

PR #110, opened 2022-08-17
feat(bivariate-color-manager) : 10843 - implementation of a map

propakov:
https://kontur.fibery.io/Tasks/Task/View-colored-bivariate-map-(Color-legends-section)-dn-fe-10843
Screenshot 2022-08-22 at 17 41 14

### DIFF
--- a/src/core/logical_layers/utils/logicalLayerFabric.ts
+++ b/src/core/logical_layers/utils/logicalLayerFabric.ts
@@ -43,6 +43,7 @@ export function createLogicalLayerAtom(
   id: string,
   renderer: LogicalLayerRenderer,
   registry: LayerRegistryAtom,
+  customMap?: maplibregl.Map | null,
 ) {
   let hasBeenDestroyed = false;
   const logicalLayerAtom = createAtom(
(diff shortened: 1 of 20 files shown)


src/core/logical_layers/utils/logicalLayerFabric.ts
```diff
@@ -43,6 +43,7 @@ export function createLogicalLayerAtom(
   id: string,
   renderer: LogicalLayerRenderer,
   registry: LayerRegistryAtom,
+  mapSource?: maplibregl.Map | null,
```
Akiyamka: maplibregl.Map in definitely not a map.source

src/core/logical_layers/utils/logicalLayerFabric.ts
```diff
@@ -75,7 +76,8 @@ export function createLogicalLayerAtom(
       },
     ) => {
       const actions: Action[] = [];
-      const map = getUnlistedState<ApplicationMap | null>(currentMapAtom);
+      const map =
```
Akiyamka, line 79: Mm, this is not at all how the map instance was supposed to be used.
When I designed the structure I assumed that we always have only one card, but here we have the existence of two cards at the same time. I'm not sure how this would work in other places where we still refer to currentMap

propakov: as we discussed I'll create an issue about refactoring and then we'll refactor this place

andrei-kucherov (approved):

PR #113, opened 2022-08-19
feat(app): static images have wrong urls after migration on go

Akiyamka:
- Loader used for image import
- Also fe 404 redirect added in router


propakov (approved):
👍

PR #114, opened 2022-08-22
fix(layers_in_area): 11069 - remove global layers /search because we don't have a separate endpoint yet

IlyaIzr:
While removing the search for global layers i've found out there's no way for us to know if user wants to see map only using link. So i stored this and when the link has map but no events - request for current "global layers" with empty geometry still sent (to at least be able to see bivariate layers)

### DIFF
--- a/src/core/shared_state/currentEvent.ts
+++ b/src/core/shared_state/currentEvent.ts
@@ -32,3 +32,4 @@ export const currentEventAtom = createAtom(
 
 export const scheduledAutoSelect = createBooleanAtom(false);
 export const scheduledAutoFocus = createBooleanAtom(false);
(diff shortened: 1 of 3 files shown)


Akiyamka (changes requested):
Can we done this without storing negative flags?

IlyaIzr:
> Can we done this without storing negative flags?

I thought about it, didn't come up with anything. 
There's a specific case in terms of `/search` for global layers - when link with coordinates without event was used. Should we really avoid negative naming for that case? If so, any ideas maybe?

IlyaIzr:
Closing PR as we dicided this logic should not be changed

PR #116, opened 2022-08-23
feat(i18n): 11243 - all translations around DN are replaced with tokens

propakov:
https://kontur.fibery.io/Tasks/Task/for-all-translations-replace-it-to-tokens-11243

### DIFF
--- a/src/core/localization/translations/en/common-messages.json
+++ b/src/core/localization/translations/en/common-messages.json
@@ -1,19 +1,180 @@
 {
   "common": {
-    "Request Timeout": "Request Timeout!",
-    "No administrative boundaries": "No administrative boundaries here!",
-    "Self intersections are not supported": "Self intersections are not supported!",
+    "km": "km",
+    "m": "m",
+    "to": "to",
 
-    "bivariate.matrix.icon.population": "Value divided by Population",
-    "bivariate.matrix.icon.area_km2": "Value divided by Area",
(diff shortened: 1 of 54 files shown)


src/core/localization/translations/en/common-messages.json
```diff
@@ -1,19 +1,180 @@
...
-    "bivariate.legend.high": "High",
-    "bivariate.legend.low": "Low",
+    "disasters": "Disasters",
+    "loading": "Loading...",
+    "loading_events": "Loading Events",
+    "about": "About",
+    "to_main_page": "to main page",
+    "legend": "Legend",
+    "maps": "maps",
+    "vertical_direction": "Vertical Direction",
```
IlyaIzr: There's theoretical concern about leading capital letters. I think it wouldn't be actual problem as we don't have so much text elements, but can you confirm it's okay?

andrei-kucherov (approved):

src/core/localization/translations/en/common-messages.json
propakov: I think it's okay, but anyway, I noticed it is not capital on designs so I turned it back noncapital.

propakov: nice catch @IlyaIzr ! I've rechecked the strings one more time.

IlyaIzr (approved):

PR #125, opened 2022-08-26
feat(toolbar): 11731 - style toolbar and fix it's buttons

IlyaIzr:
Buttons moved from sidebar are now in a new location
image
Later events panel icon will be moved from this new set of buttons. Also new tooltip selection will be dev'ed later

Also i didn't like stylelint suggestions for my .css. May be we should skip `no-descending-specificity` rule?

### DIFF
--- a/src/features/toolbar/components/Toolbar.tsx
+++ b/src/features/toolbar/components/Toolbar.tsx
@@ -31,7 +31,7 @@ export function Toolbar() {
   return (
     <div className={s.toolbar}>
       {sortByPredefinedOrder(Object.values(controls), controlsOrder).map((control) => (
-        <div key={nanoid(4)} className={s.sideBarContainer}>
+        <div key={control.id} className={s.sideBarContainer}>
           <div
             className={s.buttonWrap}
             onClick={() => control.onClick && control.onClick(!control.active)}
--- a/src/features/toolbar/components/Toolbar.tsx
+++ b/src/features/toolbar/components/Toolbar.tsx
@@ -1,5 +1,5 @@
 import { useAction, useAtom } from '@reatom/react';
-import { ActionsBar, ActionsBarBTN } from '@konturio/ui-kit';
+import { Button } from '@konturio/ui-kit';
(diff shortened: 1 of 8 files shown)


src/features/toolbar/components/Toolbar.tsx
```diff
@@ -10,23 +10,26 @@ import s from './Toolbar.module.css';
...
   function onMouseLeave() {
-    resetTooltip();
+    // resetTooltip();
   }
 
   return (
-    <ActionsBar>
+    <div className={s.toolbar}>
       {sortByPredefinedOrder(Object.values(controls), controlsOrder).map((control) => (
         <div key={nanoid(4)} className={s.sideBarContainer}>
```
propakov: i've just noticed you can use control.id instead of nanoid(4) in key

src/views/Main/Main.tsx
```diff
@@ -75,29 +72,24 @@ export function MainView({ userModel }: MainViewProps) {
       );
     }
     if (userModel?.hasFeature(AppFeature.FOCUSED_GEOMETRY_LAYER)) {
-      import('~features/focused_geometry_layer').then(
-        ({ initFocusedGeometryLayer }) => initFocusedGeometryLayer(),
+      import('~features/focused_geometry_layer').then(({ initFocusedGeometryLayer }) =>
+        initFocusedGeometryLayer(),
       );
     }
-    if (userModel?.hasFeature(AppFeature.REPORTS)) {
```
propakov, line 82: it looks like now we init report independent from AppFeature.REPORTS

`
export function initModes() {
  registerMapMode(modesControlsAtom);
  registerReportsMode(modesControlsAtom);
}`

is it expected?

propakov (approved):
LGTM.
pls consider 2 little comments

src/views/Main/Main.tsx
IlyaIzr: Yes, feature flag was responsible for reports icon only

Theese mode register functions eventually provides icons too, but in the next PR i will work with paths and think about them

PR #131, opened 2022-08-30
fix(event_list): 11728 - Make event list panel same as other panels and move in under the Analytics panel

IlyaIzr:
Panel were moved according to new design
image
image
image


Komzpa (approved):
Screenshots LGTM

propakov (approved):

PR #143, opened 2022-09-01
feat(bivariate-matrix): #11614 add bivariate matrix greeting message

andrei-kucherov:
image


IlyaIzr:
It's very wierd we have a component for 3568 lines 
image
Can we separate statics into different file?

src/features/bivariate_manager/components/BivariateGreetings/BivariateGreetingsContainer.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  useEffect(() => {
+    const disabledCookie = getCookie(GREETINGS_DISABLED_COOKIE);
+    if (!disabledCookie && !isComponentShown) {
+      setComponentShown(true);
+    }
+  });
+
+  const onCloseBtnClick = () => {
+    setComponentShown(false);
```
IlyaIzr, line 28: It's unintuitive to read for me. Why if user pushes the close button component set as not shown? 
As i understand here's the dialog window that can start showing the component. So if close button pushed, we just don't change default state - component wasn't shown, no?

propakov:
> It's very wierd we have a component for 3568 lines image Can we separate statics into different file?

It scares me. I remember we discussed it should be gif file.

src/features/bivariate_manager/components/BivariateGreetings/BivariateGreetingsContainer.tsx
```diff
@@ -0,0 +1,52 @@
...
+    document.cookie = `${GREETINGS_DISABLED_COOKIE}=true; expires=Fri, 31 Dec 9999 23:59:59 GMT`;
```
propakov: why cookie not just localStorage?

andrei-kucherov: Agree, localhost is a good alternative, Used cookie since such condition was specified in task description

src/features/bivariate_manager/components/BivariateGreetings/BivariateGreetingsContainer.tsx
andrei-kucherov: Internal state hides the whole **BivariateGreetingsContainer** component not just close button. So the logic here is: user preseses close btn -> internal component state gets chose to "not shown"  -> component hides itstelf (starts to ouput null instead of greetings element)

propakov (approved):

PR #153, opened 2022-09-09
fix(icons): 11778 - update icons according to new design

IlyaIzr:
New icons updated and also made sure old icons with enabled "currenColor" work
image
image
image


propakov (approved):
🚀

PR #159, opened 2022-09-12
feat(episodes-timline): 11912 add episodes timeline feature

Akiyamka:
image

### DIFF
--- a/src/features/event_episodes/components/EpisodesTimeline/EpisodesTimeline.tsx
+++ b/src/features/event_episodes/components/EpisodesTimeline/EpisodesTimeline.tsx
@@ -1,0 +1,72 @@
+import { Timeline } from '@konturio/ui-kit';
+import { useAtom } from '@reatom/react';
+import { useCallback, useMemo } from 'react';
+import { eventEpisodesController } from '../../controller';
+import { eventEpisodesModel } from '../../model';
+import type { TimelineProps } from '@konturio/ui-kit';
+import type { Episode } from '~core/types';
+
+type WithRequired<T, K extends keyof T> = T & { [P in K]-?: T[P] };
+
+export function EpisodesTimeline({ episodes }: { episodes: Episode[] }) {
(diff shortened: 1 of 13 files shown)


Komzpa:
an image is worth a thousand words

src/features/event_episodes/components/EpisodesTimeline/EpisodesTimeline.tsx
```diff
@@ -0,0 +1,72 @@
+import { Timeline } from '@konturio/ui-kit';
+import { useAtom } from '@reatom/react';
+import { useCallback, useMemo } from 'react';
+import { eventEpisodesController } from '../../controller';
+import { eventEpisodesModel } from '../../model';
+import type { TimelineProps } from '@konturio/ui-kit';
+import type { Episode } from '~core/types';
+
+type WithRequired<T, K extends keyof T> = T & { [P in K]-?: T[P] };
```
propakov, line 9: `-?` would be awesome to get some explanation or link 😄

propakov (approved):
👍

src/features/event_episodes/components/EpisodesTimeline/EpisodesTimeline.tsx
Akiyamka: this is literally what it looks like, it removes the optionality from the parameter (remove "?")
https://www.typescriptlang.org/docs/handbook/2/mapped-types.html#mapping-modifiers

propakov: wow, cool! thanks for explanation 🙏

src/utils/atoms/createController.ts
```diff
@@ -0,0 +1,22 @@
...
+      }
+    };
+    return acc;
+  }, {} as T);
+}
+
+const example = createFeatureController({
+  foo: () => null,
+  bar: (a: string) => a,
+});
```
IlyaIzr, line 22: should be deleted?

PR #175, opened 2022-09-23
ideas of how to fix panels

propakov:


src/views/Main/Main.module.css
```diff
@@ -123,6 +123,13 @@
   flex: 0 0 auto;
   gap: var(--unit);
   grid-column: 4;
+
+  /* what if feature turned off */
+  & > div:nth-child(2) {
```
propakov: maybe not that bad, as legend will always be visible

IlyaIzr (approved):

PR #177, opened 2022-09-26
feat(core): 12548 update typescript version yo 4.8.3

Akiyamka:


src/utils/map/cameraForGeometry.ts
```diff
@@ -2,7 +2,7 @@ import turfBbox from '@turf/bbox';
 import type { PaddingOptions } from 'maplibre-gl';
 
 function getPaddings(): PaddingOptions {
-  const width = window.visualViewport.width;
+  const width = window.visualViewport?.width ?? Infinity;
```
IlyaIzr, line 5: i wonder why infinity and not, say, `0`?

IlyaIzr (approved):

src/utils/map/cameraForGeometry.ts
Akiyamka: If guess we should show desktop layout (last return) in case we don't known screen width

PR #180, opened 2022-09-26
schedule state change, fix deps pathes

vkozio:


src/features/reports/atoms/reportResource.ts
```diff
@@ -6,7 +6,7 @@ import type { Report } from './reportsAtom';
 export const currentReportAtom = createAtom(
   { setReport: (report: Report) => report },
   ({ onAction }, state: Report | null = null) => {
-    onAction('setReport', (report) => (state = report));
+    onAction('setReport', (report) => (state = { ...report }));
     return state;
```
Akiyamka, line 10: @IlyaIzr do not mutate state, create new one

Akiyamka (approved):

PR #184, opened 2022-09-27
fix(tooltip) - added predifined position and fixed logic error when finding position

IlyaIzr:


Akiyamka (changes requested):
remove vite.config,ts.js file and add this in gitignore

vite.config.ts.js
```diff
@@ -0,0 +1,236 @@
+var __defProp = Object.defineProperty;
```
propakov: i believe this one shouldn't be commited

propakov (approved):

Akiyamka (approved):

PR #186, opened 2022-09-27
profile mode registration

IlyaIzr:


IlyaIzr:
it was an example, i will provide another PR once ready

PR #190, opened 2022-09-28
fix(header): 11720-remove-about-link-from-header

vkozio:


IlyaIzr:
I'd expect to do that _after_ about mode is implemented

src/RoutesWrap.tsx
```diff
@@ -41,19 +41,7 @@ type CommonRoutesFeaturesProps = {
...
 const afterChatContent = (loginFeature: boolean) => {
-  return (
-    <div className={s.afterChatContentWrap}>
-      <a
-        href="https://www.kontur.io/portfolio/disaster-ninja/"
-        target="_blank"
-        className={s.link}
-        rel="noreferrer"
-      >
-        <Text type="heading-m">{i18n.t('about')}</Text>
```
propakov, line 52: if you remove translations - please also remove them from `src/core/localization/translations/en/common-messages.json`

src/RoutesWrap.tsx
```diff
@@ -41,19 +41,7 @@ type CommonRoutesFeaturesProps = {
 };
 
 const afterChatContent = (loginFeature: boolean) => {
-  return (
-    <div className={s.afterChatContentWrap}>
-      <a
-        href="https://www.kontur.io/portfolio/disaster-ninja/"
-        target="_blank"
-        className={s.link}
```
propakov, line 49: .link class from Main.module.css can be removed

src/RoutesWrap.tsx
propakov: en and es both

PR #192, opened 2022-09-28
fix(panels): 12585 - add minimal height for each panel

IlyaIzr:


vkozio (approved):

src/features/layers_panel/components/MapLayersPanel/MapLayersPanel.module.css
```diff
@@ -17,7 +17,7 @@
   /* allow resize second child which is content wrap */
```
Akiyamka, line 17: any change of structure broke this, what part resizable and max height should be porps

Akiyamka (changes requested):
move min-height and resizable flag to component props

IlyaIzr:
closed as the effects will be achieved in another PR https://github.com/konturio/disaster-ninja-fe/pull/201

PR #193, opened 2022-09-28
feat(modes): 12444,11720 create landing screen mode, remove about link from header

vkozio:
[Create-landing-screen-mode-12444](https://kontur.fibery.io/Tasks/Task/Create-landing-screen-mode-12444)
[Remove-About-link-from-header-11720](https://kontur.fibery.io/Tasks/Task/Remove-About-link-from-header-11720)

image


src/views/About/About.tsx
```diff
@@ -0,0 +1,83 @@
...
+  return <AboutText />;
+}
+
+function AboutText() {
+  return (
+    <div className={s.mainWrap}>
+      <div className={s.content}>
+        <h1 className={s.pageTitle}>Welcome to Disaster Ninja!</h1>
+        <p>
+          Do you want to be notified about ongoing disasters? Are you interested in
```
propakov: do we need to have i18n here?

src/core/localization/translations/en/common-messages.json
```diff
@@ -10,7 +10,6 @@
     "disasters": "Disasters",
     "loading": "Loading...",
     "loading_events": "Loading events",
-    "about": "About",
```
propakov, line 13: pls remove it in es file also, we need to keep them the same

src/Routes.tsx
```diff
@@ -14,12 +14,41 @@ import { CommonRoutesFeatures } from './RoutesWrap';
...
 );
 
 export function RoutedApp() {
   const [{ data: userModel, loading }] = useAtom(userResourceAtom);
+
+  function isFirstTimeVisit() {
+    const firstTimeVisit =
+      userModel &&
+      !loading &&
+      location.pathname == '/' &&
```
propakov: it feels like you need to use config.baseUrl here instead of '/', when you deploy it will be /active instead of / comparing to localhost

src/Routes.tsx
```diff
@@ -14,12 +14,41 @@ import { CommonRoutesFeatures } from './RoutesWrap';
...
+      location.search == '' &&
+      !localStorage.getItem('landed');
+    if (firstTimeVisit) {
+      localStorage.setItem('landed', 'yes');
```
propakov: i know that it doesn't matter what string to store there, but 'true'  would follow our previous flags

src/Routes.tsx
propakov:
```suggestion
      location.pathname === '/' &&
```

src/Routes.tsx
```diff
@@ -14,12 +14,41 @@ import { CommonRoutesFeatures } from './RoutesWrap';
...
 
 export function RoutedApp() {
   const [{ data: userModel, loading }] = useAtom(userResourceAtom);
+
+  function isFirstTimeVisit() {
+    const firstTimeVisit =
+      userModel &&
+      !loading &&
+      location.pathname == '/' &&
+      location.search == '' &&
```
propakov:
```suggestion
      location.search === '' &&
```

src/core/app_config/appRoutes.ts
```diff
@@ -7,12 +7,14 @@ export const APP_ROUTES = {
...
 
-export function findCurrentMode(path: string) {
+export function findCurrentMode(path: string): ApplicationMode {
   let mode: ApplicationMode = 'map';
   if (path.indexOf('event') > -1) mode = 'event';
   if (path.indexOf('reports') > -1) mode = 'reports';
   if (path.indexOf('bivariate-manager') > -1) mode = 'bivariateManager';
+  if (path.indexOf('about') > -1) mode = 'about';
   return mode;
 }
```
propakov:
```suggestion
export function findCurrentMode(path: string): ApplicationMode {
  if (path.includes('event')) return 'event';
  if (path.includes('reports')) return 'reports';
  if (path.includes('bivariate-manager')) return 'bivariateManager';
  if (path.includes('about')) return 'about';
  return 'map';
}
```

src/Routes.tsx
```diff
@@ -14,12 +14,41 @@ import { CommonRoutesFeatures } from './RoutesWrap';
...
+    }
+    return false;
+  }
+
+  useEffect(() => {
+    // redirect first-time visitor "/" -> "/about"
+    if (!isFirstTimeVisit()) {
+      return;
+    }
+    const timeoutId = setTimeout(() => {
```
propakov: why do you need timeout there?

src/Routes.tsx
```diff
@@ -14,12 +14,41 @@ import { CommonRoutesFeatures } from './RoutesWrap';
...
 const { ReportPage } = lazily(() => import('~views/Report/Report'));
+const { AboutPage } = lazily(() => import('~views/About/About'));
 const { BivariateManagerPage } = lazily(
   () => import('~views/BivariateManager/BivariateManager'),
 );
 
 export function RoutedApp() {
   const [{ data: userModel, loading }] = useAtom(userResourceAtom);
+
+  function isFirstTimeVisit() {
```
propakov: function name is like it's just checker, but it mutates, I would expect this function to be clean, just return boolean, and sideeffect logic in useEffect

src/views/About/About.tsx
Komzpa: yes

src/Routes.tsx
propakov: if it's done to wait for a userModel loaded, you just need to sync dependencies
this is working without  timeouts

```
  useEffect(() => {
    const isFirstTimeVisit = () => {
      const firstTimeVisit =
        userModel &&
        !loading &&
        location.pathname === '/' &&
        location.search === '' &&
        !localStorage.getItem('landed');
      if (firstTimeVisit) {
        localStorage.setItem('landed', 'true');
        return true;
      }
      return false;
    };

    // redirect first-time visitor "/" -> "/about"
    if (!isFirstTimeVisit()) {
      return;
    }
    history.push(APP_ROUTES.about);
  }, [loading, userModel]);
```

propakov: the shorter version

  ```
useEffect(() => {
    const isFirstTimeVisit = () =>
      userModel &&
      !loading &&
      location.pathname === '/' &&
      location.search === '' &&
      !localStorage.getItem('landed');

    // redirect first-time visitor "/" -> "/about"
    if (isFirstTimeVisit()) {
      localStorage.setItem('landed', 'true');
      history.push(APP_ROUTES.about);
    }
  }, [loading, userModel]);
```

File: src/Routes.tsx (hunk 3 above)
Comment (vkozio) [reply src/Routes.tsx]: From my understanding we expect fist-time visitor at https://disaster.ninja/ and /active is our internal redirect. Will be nice to hear @Komzpa - should we check both "/" and "/active"(config.baseUrl)?

Comment (propakov) [reply src/Routes.tsx]: @vkozio /active/ added on nginx side, and js starts loading after.
You can check this if put breakpoint in disaster ninja Routes.tsx file

Comment (propakov) [reply src/Routes.tsx]: config.baseUrl is dynamic - / on locahost, /active - on prod
I believe you just need to replace location.pathname === '/' with location.pathname === config.baseUrl  and it will work

Comment (Komzpa) [reply src/Routes.tsx]: the DN2 app lives under /active/, the DN1 app lives under /live/, in the / we have just a redirector html to route to current best version.

File: src/Routes.tsx (hunk 7 above)
Comment (vkozio) [reply src/Routes.tsx]: yes, this looks enough, thanks a lot for suggestions

File: src/views/About/About.tsx [hunk 9]
```diff
@@ -0,0 +1,98 @@
...
+              the nearest fire station. Use the color legend to assess which areas require
+              attention. <br />
+              Hint: in general, green indicates low risk / few gaps, red — high risk /
+              many gaps.
+            </Trans>
+          </p>
+        </blockquote>
+
+        <p>
+          <Trans i18nKey="about.p5">
```
propakov: why to use  <Trans  component there if it's only a line of text without dividers? 
same for `<Trans i18nKey="about.p2">`

propakov (approved):
LGTM. Pls consider one question(about `<Trans>` component), maybe it's a point to improve.

src/views/About/About.tsx
propakov: I've replaced it with i18n.t() and don't see any difference.

IlyaIzr:
All those keys to english file will be added to spanish localization later?
image

IlyaIzr (approved):

propakov:
> All those keys to english file will be added to spanish localization later? image

@IlyaIzr I believe yes. We just add keys to en file and when we will have much of them - send to translators.

PR #194, opened 2022-09-28
fix(sidebar): 12440-change-reports-icon-to-suggest-it-is-problems-reports

vkozio:
image

[12440](https://kontur.fibery.io/Tasks/Task/Change-reports-icon-to-suggest-it-is-problems-reports-12440)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "disaster-ninja-fe",
-  "version": "2.1.2",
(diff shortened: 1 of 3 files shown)


propakov (approved):
👍 
we also decided that it's a good practise to add at least 1 screenshot in PR for clarity

propakov:
and supercool will be adding a link to task  from fibery :D

PR #201, opened 2022-10-05
chore(layouts): 12585, 12580, 12588, 12590, 12607 - implement dynamic layouts

IlyaIzr:
1920
image
Laptop
image
Mobile
image
image
image


Komzpa:
screenshots LGTM. the fitBounds behavior is not shown on the screenshots.

src/views/Main/Layouts/Desktop/Desktop.module.css
```diff
@@ -0,0 +1,59 @@
...
+  gap: var(--unit);
+  padding: var(--unit);
+
+  /* analytics+layers column, advanced analytics column, map column, layers&legends column */
+  grid-template-columns:
+    minmax(320px, 1fr) fit-content(400px) minmax(440px, 10fr)
+    minmax(320px, 1fr);
+  grid-template-rows: calc(100% - 40px) 40px;
+}
+/* .firstColumn, .secondColumn, .thirdColumn, .fourthColumn{
```
propakov: why to save this comment? can't find it in code or comments

propakov (commented):
From code side it LGTM, but I've found some bugs while testing.
1) resizing corners are a bit different
image
2) scroll bar on mobile view goes out of screen, and it's not possible to scroll to the bottom of the list
near the bottom
image
the most bottom position
image

src/core/store/columnContext.tsx
```diff
@@ -0,0 +1,121 @@
+import { useContext, createContext } from 'react';
+import type { MutableRefObject } from 'react';
+
+export const ColumnContext = createContext<Resizer | string>('no provider was given');
```
propakov: I also can't find a reason to have string there. We don't put there other strings, only one place here. 
Maybe just null? And there `if (typeof columnContext === 'string') return;` process as null

IlyaIzr: i was hoping it would give better readability, but since you've asked, it didn't - so i'll change it

propakov:
@IlyaIzr after fix it's double scroll 
Screenshot 2022-10-06 at 16 18 43
and icon on mobile is without shadows
image

src/core/store/columnContext.tsx
```diff
@@ -0,0 +1,121 @@
...
+    resizeObserver.observe(this.column.current);
+    this.destroy = () => resizeObserver.unobserve(this.column!.current!);
+  }
+
+  addPanel(panel: PanelMeta) {
+    return this.panels.add(panel);
+  }
+
+  destroy() {
+    // TODO
```
Akiyamka: what you want todo here?

src/views/Main/Main.tsx
```diff
@@ -1,20 +1,19 @@
...
 import type { UserDataModel } from '~core/auth';
-import type { MutableRefObject } from 'react';
 
-const { EditFeaturesOrLayerPanel } = lazily(
-  () =>
-    import(
-      '~features/create_layer/components/EditFeaturesOrLayerPanel/EditFeaturesOrLayerPanel'
-    ),
-);
+// To be restored later
```
Akiyamka: Please add some info about when in must be restored

Akiyamka:
Nice work!

Akiyamka (approved):
Code looks nice, just few comments about comments : )

PR #207, opened 2022-10-10
fix(core): 12819 default layers request use empty string as app id

Akiyamka:
## In general
In cause bug with incorrect default layers for some custom apps.
Since after refactor asyncAtom have more strict check - it stop getting default layers at all (this is how I found this bug)

## Regression - url parameters common for all routes
For fix that I remove unstable code part that restoring url search parameters after route changed.
Right now url store common for all app so url parameters will no changed depending from mode.

If we need different url stores for every mode (route) - this must be designed and implemented as separate task
(thread: https://konturio.slack.com/archives/C02DEKP2MKP/p1665419230269439)

## Breaking change - default app id and default layers not atoms anymore
No one has used those in reactive way, and at same time it much more simpler and readable as simple fetches chain 

## Linter change
Restriction for console.debug was removed

## Perf improvement
At lest one extra render removed when app booting - i remove disable action from layer that already disabled

## Chore
All modes registrators now in separate folder


src/core/modes/registrations/mapMode.tsx
```diff
@@ -12,8 +12,5 @@ export function registerMapMode(modesControlAtom: ModesControlsAtom) {
     onClick() {
       currentModeAtom.setCurrentMode.dispatch('map');
     },
-    onChange(isActive) {
```
propakov, line 15: it looks like we need to remove this code section in all mode files to be consistent 
src/core/modes/registrations

propakov (approved):
🚀

src/core/modes/registrations/mapMode.tsx
Akiyamka: agree, removed

PR #214, opened 2022-10-12
Hotfix(landing)  12444 create landing screen mode

vkozio:
open external links in new window, show email on page, go to map link works via route (preserving map state)


IlyaIzr (approved):
couldn't checkout locally but code looks okay 
image

PR #221, opened 2022-10-13
refactor(areaLayers): 11273 - optimizing requests and calculations

Akiyamka:
- createResourseAtom in areaLayers was repalced with createAsyncAtom
- new endpoints used to get layers lists. Now we get static global layers and dynamic area layers separately. It significantly improve first layers list loading time, and simplify logic
- area layers requests now canceled if user change his mind while they fetched
- layers details now have cache - so we get only layers details that was never requested before. Cache will reset when user or application changed. Event layer never cached 
- eventShape excluded from cache, because it depends from event and changes after event changed.


IlyaIzr:
Here's the bug: 
Enable non-global layers for event\disaster
Switch disaster
Non-global layers didn't update on the map and also no additional `/details` request was created 
image

src/features/layers_in_area/atoms/areaLayersDetailsResource.ts
```diff
@@ -15,46 +20,72 @@ export interface DetailsRequestParams {
...
+    onAction('add', (newData) => {
+      state = new Map(state);
+      newData.forEach((details) => {
+        state.set(details.id, details);
+      });
+    });
+    return state;
+  },
+  'areaLayersDetailsResourceAtomCache',
+);
```
IlyaIzr: how about moving it to separate file? this file is getting pretty big

src/features/layers_in_area/atoms/layersInAreaAndEventLayerResource.ts
```diff
@@ -0,0 +1,75 @@
...
+    return removeEmpty({
+      appId,
+      eventFeed,
+      eventId,
+      geoJSON,
+    });
+  },
+);
+
+export const layersInAreaAndEventLayerResource = createAsyncAtom(
```
IlyaIzr, line 59: Why do we need to add that EventLayer is here too? considering we're getting the layers related to event-id and event-feed-id?

src/features/layers_in_area/atoms/layersInAreaAndEventLayerResource.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (appId === null) return null;
+
+    const geoJSON = focusedGeometry.geometry;
+    const eventId = getEventId(focusedGeometry);
+    // getUnlistedState for avoid extra request when layer if
+    // Updates flow:
+    // 1. eventId changed (ignoring)
+    // 2. focusedGeometry start loading for new event
+    // 3. focusedGeometry loaded
+    // 4. get layers for new geometry
```
IlyaIzr, line 47: That general logic comment should be higher

src/features/layers_in_area/atoms/layersInAreaAndEventLayerResource.ts
```diff
@@ -0,0 +1,75 @@
...
+    const focusedGeometry = get('focusedGeometryAtom');
+    const appId = get('currentApplicationAtom');
+
+    // Check required
+    if (focusedGeometry === null) return null;
+    if (appId === null) return null;
+
+    const geoJSON = focusedGeometry.geometry;
+    const eventId = getEventId(focusedGeometry);
+    // getUnlistedState for avoid extra request when layer if
```
IlyaIzr: and this comment should beright above `getUnlistedState` and you should fix the text a bit something like 
```suggestion
    // getUnlistedState used to avoid extra request for the same event in different feeds
```

If i've understood correctly. Anyway it needs fixing

src/utils/atoms/combineAtoms.ts
```diff
@@ -13,3 +14,58 @@ export function combineAtoms<T extends AtomsMap>(
     }, {} as { [key in keyof T]: T[key] extends Atom<infer S> ? S : unknown });
   });
 }
+
+type AsyncAtomsMap = { [key: string]: AsyncAnyAtom<AtomBinded<any>> };
+type CombinedAsyncAtomsState<T> = {
+  loading: boolean;
+  error:
+    | {
```
IlyaIzr, line 22:
```suggestion
    {
```

src/utils/atoms/combineAtoms.ts
```diff
@@ -13,3 +14,58 @@ export function combineAtoms<T extends AtomsMap>(
...
+    return Object.keys(shape).reduce(
+      (acc, key: keyof T) => {
+        const { loading, error, data, lastParams } = get(key) as AsyncAtomState<
+          unknown,
+          Record<string, unknown>
+        >;
+        if (loading) acc.loading = true;
+        if (error) {
+          if (acc.error === null) {
+            // @ts-ignore
```
IlyaIzr, line 52: ts ignores are expected in this file, right?

src/utils/atoms/combineAtoms.ts
```diff
@@ -13,3 +14,58 @@ export function combineAtoms<T extends AtomsMap>(
...
+    [key in keyof T]: T[key] extends AsyncAnyAtom<infer P, infer F>
+      ? AtomState<T[key]>['data']
+      : unknown;
+  };
+  lastParams: {
+    [key in keyof T]: T[key] extends AsyncAnyAtom<infer P, infer F>
+      ? AtomState<T[key]>['lastParams']
+      : unknown;
+  };
+  dirty: true;
```
IlyaIzr, line 38: I'd like to have an explanation for `dirty`. There's no explanation for it in `createAsyncAtom` too

src/utils/atoms/combineAtoms.ts
```diff
@@ -13,3 +14,58 @@ export function combineAtoms<T extends AtomsMap>(
...
+          }
+        }
+        // @ts-ignore
+        acc.lastParams[key] = lastParams;
+        // @ts-ignore
+        acc.data[key] = data;
+
+        return acc;
+      },
+      { dirty: true.valueOf, data: {} } as unknown as CombinedAsyncAtomsState<T>,
```
IlyaIzr:
```suggestion
      { dirty: true, data: {} } as unknown as CombinedAsyncAtomsState<T>,
```

or am I missing something...

src/features/layers_in_area/atoms/layersInAreaAndEventLayerResource.ts
Akiyamka: because it true. 
We get 1) layers in area and 2) event layer here (in ui in named as "event shape". In near feature we will get event shape from other api in other feature

src/features/layers_in_area/atoms/layersInAreaAndEventLayerResource.ts
Akiyamka: in explain `getUnlistedState` under comment. I fix text a little bit

src/features/layers_in_area/atoms/layersInAreaAndEventLayerResource.ts
Akiyamka: While commend block is one big explanation, but first line had mistake,  fixed

src/utils/atoms/combineAtoms.ts
Akiyamka: it is prettier formatting

src/utils/atoms/combineAtoms.ts
Akiyamka: unfortunately I'm not found a way to fix that without ts-ignore. If you now how, please told me

src/utils/atoms/combineAtoms.ts
Akiyamka: hmm, it's pretty standard thing
https://en.wikipedia.org/wiki/Dirty_bit
https://en.wikipedia.org/wiki/Cache_(computing)#Dirty
And it have comment
image

src/utils/atoms/combineAtoms.ts
Akiyamka: nice catch!

IlyaIzr (approved):

PR #223, opened 2022-10-14
feat(modes): 12288, 12994, 12396, 12292 - create user profile mode

IlyaIzr:
Several settings to change are implemented at this stage
Component of <Textarea /> will be created once i figure out my troubles with ui-repo
ALso select component needs change to have labels\placeholders on top

Ability to push save button is only provided when user changed their settings - according to requireonments

image


vkozio (approved):
I'd suggest using CSS3 columns, however I can add it in my PR

PR #225, opened 2022-10-15
feat(core): 11953 improve development experience for vite linking to dn

Akiyamka:
image

### DIFF
--- a/scripts/auto-link/config.mjs
+++ b/scripts/auto-link/config.mjs
@@ -10,7 +10,7 @@ export async function geConfig(pathToConfig = './scripts/auto-link/.path') {
     return { pathToUIKitRepo };
   } catch(e) {
     log.debug(e)
-    const pathToUIKitRepo = await prompts.askPathToFile('Path to folder with @kontur/ui repo');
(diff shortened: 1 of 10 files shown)


scripts/auto-link/config.mjs
```diff
@@ -0,0 +1,17 @@
+import { readFileSync, writeFileSync } from "fs";
+import prompts from './prompts.mjs';
+import { logger } from './logger.mjs';
+
+const log = logger('Config reader')
+// @ts-check
```
IlyaIzr, line 6:
```suggestion

```

IlyaIzr:
[Winows 11]
So i've tried absolute path to ui kit (with win slashes and backslashes also)
We could improve error message
Also despite path being unreadable the path got saved so next time i run script - it just throws error again (i'd suggest not saving it if path is not OK)
image
image
relative path `../ui` also didn't work for me

Tried both on powershell and bash

scripts/auto-link/config.mjs
Akiyamka: Why you suggest delete this ?

IlyaIzr: Isn't it checked by ts anyway?

Akiyamka: it checked in vscode

PR #227, opened 2022-10-19
feat: 12826 Unify deployment process. Update docker tags and publish …

azapasnik:
…trigger

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,7 +86,8 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_REPO }}
           tags: |
-            type=raw,value=${{ github.ref_name }}.{{sha}}.${{ github.run_attempt }}
+            type=raw,value=${{ github.ref_name }}.{{sha}}.${{ github.run_attempt }},enable=${{github.event_name == 'push' || github.event_name == 'workflow_dispatch'}}
+            type=raw,value=${{ github.head_ref }}.{{sha}}.${{ github.run_attempt }},enable=${{github.event_name == 'pull_request'}}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -62,7 +62,6 @@ jobs:
 
   publish:
     needs: build
-    if: ${{ github.event_name == 'workflow_dispatch' || (github.event_name == 'push' && startsWith(github.ref_name, 'release')) }}
```
Akiyamka, line 65: This was done on purpose, need to keep

Akiyamka (changes requested):
auto publishing must happens only for releases

.github/workflows/build.yml
azapasnik: what is the purpose?

azapasnik:
Why to autopublish only releases?
I'd like to keep the same approach for all projects

.github/workflows/build.yml
Akiyamka: it was decided to add this restriction because without it we had too many Images and it made it hard to find the right one.
Given that we do not use these images (only release and sometimes main), it is logical to remove the extra action, and trigger it manually for rare cases when it needed

azapasnik: but this rule still exists
```
on:
  push:
    branches:
      - 'main'
      - 'release-*'
  pull_request:
    types:
      - opened
      - reopened
      - ready_for_review
      - synchronize
 ```
so it will not be triggered for dev branches

Akiyamka:
> Why to autopublish only releases? I'd like to keep the same approach for all projects  

I don't mind autopublish only releases in all projects =)

Akiyamka (approved):

PR #230, opened 2022-10-21
fix(reports): 13006 - fix report page styles

IlyaIzr:
Scroll bar is now at it's place
Sticky header is visible
Paddings fixed for laptops
image
image
stick bar fixed for big screens (> 1920px width)
image

### DIFF
--- a/src/features/reports/atoms/tableAtom.ts
+++ b/src/features/reports/atoms/tableAtom.ts
@@ -77,7 +77,9 @@ export const tableAtom = createAtom(
       });
     });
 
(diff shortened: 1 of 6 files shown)


Komzpa (approved):
Looks good in general. Have a feeling we need to somehow move the breakpoints pixel values into variables so maybe this needs a newbbacklog task.

IlyaIzr:
Yes breakpoints would be cool but hard to implement to ATM, as far as i remember (we discussed this few months ago)

PR #232, opened 2022-10-21
fix(repots, analytics, advanced_analytics): 12813 - move reports and pick disaster text under feature flags

IlyaIzr:
we need this to create Smart City version of DN


Komzpa (approved):
LGTM

PR #237, opened 2022-10-27
fix(core): 13025 dn2-doesn-t-load-after-some-default-layers-are-disabled

vkozio:
https://kontur.fibery.io/Tasks/Task/DN2-doesn't-load-after-some-default-layers-are-disabled-13025

### DIFF
--- a/src/Routes.tsx
+++ b/src/Routes.tsx
@@ -7,7 +7,6 @@ import { OriginalLogo } from '~components/KonturLogo/KonturLogo';
 import history from '~core/history';
 import config from '~core/app_config';
 import { userResourceAtom } from '~core/auth/atoms/userResource';
-// import { LoginForm } from '~features/user_profile';
 import { APP_ROUTES } from '~core/app_config/appRoutes';
 import { AppFeature } from '~core/auth/types';
 import s from './views/Main/Main.module.css';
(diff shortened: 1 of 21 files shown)


src/Routes.tsx
```diff
@@ -7,7 +7,7 @@ import { OriginalLogo } from '~components/KonturLogo/KonturLogo';
 import history from '~core/history';
 import config from '~core/app_config';
 import { userResourceAtom } from '~core/auth/atoms/userResource';
-import { LoginForm } from '~features/user_profile';
+// import { LoginForm } from '~features/user_profile';
```
Akiyamka: remove if not needed

src/utils/atoms/createAsyncAtom/createAsyncAtom.test.ts
```diff
@@ -407,7 +407,7 @@ describe('Resource reactivity', () => {
     });
   });
 
-  test('Refetch when deps object changed', async ({ store }) => {
+  test.todo('Refetch when deps object changed', async ({ store }) => {
```
Akiyamka, line 410: Why we disable this test?
Something broken?

src/utils/atoms/createAsyncAtom/createAsyncAtom.ts
```diff
@@ -235,7 +237,8 @@ export function createAsyncAtom<
             }
           } else {
             // Deps is primitive
-            schedule((dispatch) => dispatch(create('request', depsAtomState as any)));
+            if (depsAtomState !== null)
```
Akiyamka, line 240: Please add test for this new behavior

PR #243, opened 2022-11-01
fix(core, focused_geomtetry_editor): 12806 - change draw tools initial state

IlyaIzr:
Focused geometry editor now started with modify mode with selection of all geometry passed to it
Hint to start drawing was removed
image

This was done by creating boolean-flag atom, that activated by feature and when geometry is provided - no matter the feature ammount - they get selected
image

### DIFF
--- a/src/core/draw_tools/atoms/combinedAtom.ts
+++ b/src/core/draw_tools/atoms/combinedAtom.ts
@@ -1,3 +1,4 @@
+import _ from 'lodash';
 import { createAtom } from '~utils/atoms';
 import { currentMapAtom, currentNotificationAtom } from '~core/shared_state';
 import { setMapInteractivity } from '~utils/map/setMapInteractivity';
@@ -7,7 +8,7 @@ import { drawnGeometryAtom } from './drawnGeometryAtom';
 import { drawModeLogicalLayerAtom } from './logicalLayerAtom';
 import {
   selectedIndexesAtom,
-  selIndexesForCurrentGeometryAtom,
+  setIndexesForCurrentGeometryAtom,
 } from './selectedIndexesAtom';
 import { isDrawingStartedAtom } from './isDrawingStartedAtom';
 import { drawModeRenderer } from './logicalLayerAtom';
@@ -47,7 +48,7 @@ export const combinedAtom = createAtom(
 
     selectedIndexesAtom,
     setIndexes: (indexes: number[]) => indexes,
-    selIndexesForCurrentGeometryAtom,
+    setIndexesForCurrentGeometryAtom,
 
     setDrawingIsStarted: (isStarted: boolean) => isStarted,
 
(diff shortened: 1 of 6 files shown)


src/core/draw_tools/atoms/combinedAtom.ts
```diff
@@ -95,9 +100,17 @@ export const combinedAtom = createAtom(
...
     });
 
-    onChange('drawnGeometryAtom', (featureCollection) =>
-      (state.drawnGeometryAtom ?? []).forEach((cb) => cb(featureCollection)),
-    );
+    onChange('drawnGeometryAtom', (featureCollection, prevCollection) => {
+      if (selectCurrentGeometryWasRequested) {
+        actions.push(selIndexesForCurrentGeometryAtom.set(false));
+        const indexes: number[] = [];
+        for (let i = 0; i < featureCollection.features.length; i++) {
```
vkozio: you can use https://lodash.com/docs/4.17.15#times instead for loop and make it bit more compact

vkozio (approved):

src/core/draw_tools/atoms/combinedAtom.ts
propakov: or [...Array(featureCollection.features.length).keys()]

propakov (commented):
found a possible bug: 
- open event from event feed
- click focus on freehand geometry (you see points draggable and draw mode on)
- open another event from event feed (draw mode is still on, but geometry is from another event)
- click delete button from drawbox (it removes geometry from previous event, actual event gepmetry points are not draggable)

IlyaIzr:
> found a possible bug:
> 
> * open event from event feed
> * click focus on freehand geometry (you see points draggable and draw mode on)
> * open another event from event feed (draw mode is still on, but geometry is from another event)
> * click delete button from drawbox (it removes geometry from previous event, actual event gepmetry points are not draggable)

Since user enters draw mode - he enters it with specific geometry he works on - we didn't ever try to throw any geometry inside draw mode while this happens. Also when user pushes 'finish drawing' - he applies drawing mode state to focus geometry. If it's nullish (no features) - focus geometry should become empty.

I'll add this point to discuss on draw tools refactoring.
ALso this logic would've still be there on prev versions. before this PR

src/core/draw_tools/atoms/combinedAtom.ts
IlyaIzr: you guys clearly don't like vanilla js :D

src/core/draw_tools/atoms/combinedAtom.ts
```diff
@@ -1,11 +1,15 @@
+import _ from 'lodash';
```
propakov: i believe if you import whole lodash it won't be possible to tree shake it. 
suggest to import only needed stuff like we did before 
`import debounce from 'lodash/debounce';`

propakov (approved):
pls fix lodash import and other than that LGTM 🔥

PR #245, opened 2022-11-03
feat: 12133 Set up redirect from DN1 to DN2

azapasnik:


server/live/index.html
```diff
@@ -0,0 +1,33 @@
...
+<head>
+  <title>Disaster Ninja by kontur.io</title>
+  <meta name="description"
+        content="Disaster Ninja is a platform for assesment of mapping needs in case of natural disaster. See what's happening, what are available map sources, explore coverage quality of OpenStreetMap, chart the area of interest, export and import as geojson.">
+  <meta http-equiv="refresh" content="0; url=/active/"/>
+  <script>
+    const hash = window.location.hash.substring(1);
+    let params = {};
+    hash.split(';').map(hk => {
+      let temp = hk.split('=');
```
IlyaIzr: what does "hk" means?

IlyaIzr: also we use `let` declaration only when variable will be reassigned later, so 
```suggestion
      const temp = hk.split('=');
```

server/live/index.html
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <title>Disaster Ninja by kontur.io</title>
+  <meta name="description"
+        content="Disaster Ninja is a platform for assesment of mapping needs in case of natural disaster. See what's happening, what are available map sources, explore coverage quality of OpenStreetMap, chart the area of interest, export and import as geojson.">
+  <meta http-equiv="refresh" content="0; url=/active/"/>
+  <script>
+    const hash = window.location.hash.substring(1);
```
IlyaIzr: You can also briefly describe what that script does, for example
```suggestion
  <script>
  // This script takes url hash, transforms it to DN 2 url format and initiates reirect with those params
    const hash = window.location.hash.substring(1);
```
Or add comments according to the actions described

IlyaIzr:
still need @Akiyamka review for go part as `parsedTemplate, _ := ` is moonlanguage to me

server/live/index.html
azapasnik: just variable name. hash key

server/live/index.html
```diff
@@ -0,0 +1,34 @@
...
+    });
+    let queryUrl = "";
+    if (params.id) {
+      queryUrl += "event=" + params.id;
+    }
+    if (params.position && params.zoom) {
+      if (queryUrl !== "") {
+        queryUrl = queryUrl + "&";
+      }
+      queryUrl += "map=" + params.zoom + "/" + params.position.replace(",", "/");
```
Komzpa, line 24: @vkozio now is flipping coordinates here, synchronize

server/live/index.html
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <title>Disaster Ninja by kontur.io</title>
+  <meta name="description"
+        content="Disaster Ninja is a platform for assesment of mapping needs in case of natural disaster. See what's happening, what are available map sources, explore coverage quality of OpenStreetMap, chart the area of interest, export and import as geojson.">
+  <meta http-equiv="refresh" content="0; url=/active/"/>
```
Komzpa: I suspect this way it may happen that it works faster than JS in some cases. changing 0 -> 1 should give enough time for it to work every time.

Komzpa (approved):

PR #249, opened 2022-11-06
fix(event_list): 11841 - fix autoselect fired when event is already selected

IlyaIzr:
If url was applied with event-id not presented in event list but availible through api call - warning title is presented for 5 seconds and autoselect is not fired 
image

I also added error message for 404 event requests as there was no other way for user to know that event cannot be reached

### DIFF
--- a/src/features/current_event/atoms/currentEventResource.ts
+++ b/src/features/current_event/atoms/currentEventResource.ts
@@ -2,6 +2,7 @@ import { createAtom } from '~utils/atoms';
 import { apiClient } from '~core/apiClientInstance';
 import { currentEventAtom, currentEventFeedAtom } from '~core/shared_state';
 import { createAsyncAtom } from '~utils/atoms/createAsyncAtom';
+import { i18n } from '~core/localization';
 import type { EventWithGeometry } from '~core/types';
 
 const eventDependencyAtom = createAtom(
@@ -29,7 +30,13 @@ export const currentEventResourceAtom = createAsyncAtom(
         `/events/${deps.feed.id}/${deps.event.id}`,
         undefined,
         undefined,
-        { signal: abortController.signal, errorsConfig: { dontShowErrors: true } },
+        {
+          signal: abortController.signal,
+          errorsConfig: {
+            dontShowErrors: false,
+            messages: i18n.t('current_event.not_found_request'),
+          },
+        },
       );
       if (responseData === undefined) throw 'No data received';
       return responseData;
--- a/src/features/current_event/atoms/currentEventResource.ts
+++ b/src/features/current_event/atoms/currentEventResource.ts
@@ -34,7 +34,7 @@ export const currentEventResourceAtom = createAsyncAtom(
           signal: abortController.signal,
           errorsConfig: {
             dontShowErrors: false,
-            messages: i18n.t('current_event.not_found_request'),
+            messages: { 404: i18n.t('current_event.not_found_request') },
           },
         },
       );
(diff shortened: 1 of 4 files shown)


IlyaIzr:
i've also changed "event not found in the list" toast color from error to warning state as it's not stopping user from using app properly

src/features/events_list/atoms/autoSelectEvent.ts
```diff
@@ -23,10 +21,20 @@ export const autoSelectEvent = createAtom(
...
+          schedule((dispatch) =>
+            dispatch(
+              currentNotificationAtom.showNotification(
+                'warning',
+                { title: i18n.t('event_list.no_event_in_feed') },
+                5,
+              ),
+            ),
+          );
+        } else if (currentEventNotInNewList) {
```
propakov: i would suggest to get rid of else if logic 
like here (feels simplier)
```
      if (currentEventNotInNewList) {
          if (currentEvent?.id) {
            // This case happens when call for event by provided url id didn't return event
            schedule((dispatch) =>
              dispatch(
                currentNotificationAtom.showNotification(
                  'warning',
                  { title: i18n.t('event_list.no_event_in_feed') },
                  5,
                ),
              ),
            );
          } else {
            const firstEventInList = eventListResource.data[0];
            schedule((dispatch) => {
              dispatch([
                scheduledAutoSelect.setFalse(),
                scheduledAutoFocus.setTrue(),
                currentEventAtom.setCurrentEventId(firstEventInList.eventId),
              ]);
            });
          }
        }

File: src/features/events_list/atoms/autoSelectEvent.ts [hunk 2]
```diff
@@ -23,10 +21,20 @@ export const autoSelectEvent = createAtom(
       ) {
         const currentEvent = getUnlistedState(currentEventAtom);
         const currentEventNotInNewList =
-          eventListResource.data.findIndex(
-            (e) => e.eventId === currentEvent?.id,
-          ) === -1;
-        if (currentEventNotInNewList) {
+          eventListResource.data.findIndex((e) => e.eventId === currentEvent?.id) === -1;
```
propakov:
```suggestion
          !eventListResource.data.some((e) => e.eventId === currentEvent?.id);
```

src/features/events_list/atoms/autoSelectEvent.ts
IlyaIzr: for me it's one more nesting\intendation level

IlyaIzr: I did compromise

propakov (approved):

src/features/events_list/atoms/autoSelectEvent.ts
```diff
@@ -22,11 +20,25 @@ export const autoSelectEvent = createAtom(
...
-          eventListResource.data.findIndex(
-            (e) => e.eventId === currentEvent?.id,
-          ) === -1;
-        if (currentEventNotInNewList) {
+        const currentEventNotInTheList = !eventListResource.data.some(
+          (e) => e.eventId === currentEvent?.id,
+        );
+
+        // do nothing
+        if (!currentEventNotInTheList) return;
```
propakov: it's pretty clear why early return is needed :)
Btw, i remember we discussed on FE meetup about return undefined from atom. Is it ok?

IlyaIzr: good point, i thought it was stateless atom
i will `return state` then

Akiyamka (approved):

src/features/current_event/atoms/currentEventResource.ts
```diff
@@ -29,7 +30,13 @@ export const currentEventResourceAtom = createAsyncAtom(
         `/events/${deps.feed.id}/${deps.event.id}`,
         undefined,
         undefined,
-        { signal: abortController.signal, errorsConfig: { dontShowErrors: true } },
+        {
+          signal: abortController.signal,
+          errorsConfig: {
+            dontShowErrors: false,
```
Akiyamka, line 36: I expect now you will always see error when request canceled, that I guess not what you want

PR #250, opened 2022-11-08
feat(locate_me): 11785 - locate me feature added to toolbar

IlyaIzr:
Toolbar button
image
Triggers dialogue window
image
In case user said 'deny' or other errors - error toast shown
image
Otherwise - we zoom to location with zoom 14

### DIFF
--- a/src/features/locate_me/index.tsx
+++ b/src/features/locate_me/index.tsx
@@ -1,0 +1,50 @@
+import { Locate24 } from '@konturio/default-icons';
+import {
+  currentMapPositionAtom,
+  currentNotificationAtom,
+  toolbarControlsAtom,
+} from '~core/shared_state';
+import { controlGroup, controlVisualGroup } from '~core/shared_state/toolbarControls';
+import { i18n } from '~core/localization';
+import { LOCATE_ME_CONTROL_ID, LOCATE_ME_CONTROL_NAME } from './constants';
+
+function successCb(location: GeolocationPosition) {
+  const { coords } = location;
+  const { latitude: lat, longitude: lng } = coords;
+
+  currentMapPositionAtom.setCurrentMapPosition.dispatch({ lat, lng, zoom: 14 });
+  disableControl();
+}
+
+function errorCb(error: GeolocationPositionError) {
+  currentNotificationAtom.showNotification.dispatch(
+    'warning',
+    { title: i18n.t('locate_me.get_location_error' + ': ' + error.message) },
+    3,
+  );
+  disableControl();
+}
+
+export function initLocateMe() {
+  toolbarControlsAtom.addControl.dispatch({
+    id: LOCATE_ME_CONTROL_ID,
+    name: LOCATE_ME_CONTROL_NAME,
+    title: i18n.t('locate_me.feature_title'),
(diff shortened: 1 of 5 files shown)


src/features/locate_me/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const { latitude: lat, longitude: lng } = coords;
+
+  currentMapPositionAtom.setCurrentMapPosition.dispatch({ lat, lng, zoom: 14 });
+  disableControl();
+}
+
+function errorCb(error: GeolocationPositionError) {
+  currentNotificationAtom.showNotification.dispatch(
+    'warning',
+    { title: i18n.t('locate_me.get_location_error' + ': ' + error.message) },
```
Komzpa: https://wordpress.org/support/topic/i18n-please-dont-build-sentences-by-concatenating-substrings/

Komzpa: also, error.message gets untranslated?

propakov: it's not clear what's going on here, could you please explain? @IlyaIzr 
btw we shouldn't pass dynamic stuff to i18n.t, as it can't work with future i18n static analyser

src/features/locate_me/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+} from '~core/shared_state';
+import { controlGroup, controlVisualGroup } from '~core/shared_state/toolbarControls';
+import { i18n } from '~core/localization';
+import { LOCATE_ME_CONTROL_ID, LOCATE_ME_CONTROL_NAME } from './constants';
+
+function successCb(location: GeolocationPosition) {
+  const { coords } = location;
+  const { latitude: lat, longitude: lng } = coords;
+
+  currentMapPositionAtom.setCurrentMapPosition.dispatch({ lat, lng, zoom: 14 });
```
propakov: const MAX_ZOOM = 14 will be good? what do you think?

src/features/locate_me/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  disableControl();
+}
+
+function errorCb(error: GeolocationPositionError) {
+  currentNotificationAtom.showNotification.dispatch(
+    'warning',
+    { title: i18n.t('locate_me.get_location_error' + ': ' + error.message) },
+    3,
```
propakov, line 31: it would be good to reuse this param with const and reuse it with disableControl 3000 value

src/features/locate_me/index.tsx
IlyaIzr: Yeah, that's my bad. In fact, `error.message` actually seems enough as it provides propper error message by browser. I'll fix that

src/features/locate_me/index.tsx
IlyaIzr: Hm i can't see how showing casual error notification is connected with `disableControl`, which is action of toolbarControlsAtom

propakov: ah, right! it's about error notification, than it's not needed, sure. thanks!

propakov (approved):

PR #251, opened 2022-11-08
fix(layers_in_area): 13152 - selected area is not correct for current event

IlyaIzr:
I've simply applied cache that was already implemented but was only storing layers and stopping them from being requested

What was broken: select event A with event-related geometry layer, select event B, select event A - cached layers not applied.

### DIFF
--- a/src/features/layers_in_area/atoms/areaLayersLegendsAndSources.ts
+++ b/src/features/layers_in_area/atoms/areaLayersLegendsAndSources.ts
@@ -2,7 +2,10 @@ import { createAtom } from '~utils/atoms/createPrimitives';
 import { layersSourcesAtom } from '~core/logical_layers/atoms/layersSources';
 import { layersLegendsAtom } from '~core/logical_layers/atoms/layersLegends';
 import { legendFormatter } from '~features/layers_in_area/utils/legendFormatter';
+import { focusedGeometryAtom } from '~core/shared_state';
+import { getEvendId } from '~core/shared_state/focusedGeometry';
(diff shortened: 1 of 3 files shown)


src/features/layers_in_area/atoms/areaLayersLegendsAndSources.ts
```diff
@@ -63,6 +67,16 @@ export const areaLayersLegendsAndSources = createAtom(
...
       }, new Map<string, LayerInAreaDetails>());
 
+      // apply cached layers if any are already stored for current eventId
+      const focusedGeometry = get('focusedGeometryAtom');
+      const eventId = getEvendId(focusedGeometry);
+      if (eventId) {
+        const cachedLayers = getUnlistedState(areaLayersDetailsResourceAtomCache).get(
+          eventId,
+        );
+        cachedLayers?.forEach((layer) => layersDetailsData.set(layer.id, layer));
```
Akiyamka, line 77: The idea was that each update of the resource atom only adds new data to layersSourcesAtom and layersLegendsAtom
(83 and 97 lines of this file). So the cache in this case is only needed to compose the new request.

I didn't take into account that they are deleted somewhere else.

I think we need a test for this behavior, this bug has been on the production for some time now because it's hard to catch it by hand

Akiyamka (approved):

src/features/layers_in_area/atoms/areaLayersLegendsAndSources.ts
IlyaIzr: skipping that as it's not yet clear how to make test for atom that relied on `getUnlistedState`

PR #254, opened 2022-11-09
feat(event_episodes): add delay option for loading state

Akiyamka:
createStateMap fabric now can accept settings with delay for showing loading state


propakov (approved):
👍

PR #256, opened 2022-11-10
feat(i18n): 12862 - gettext format support on frontend side added

propakov:
https://kontur.fibery.io/Tasks/Task/Add-gettext-format-support-on-frontend-side-12862

Screenshot 2022-11-11 at 10 17 06


vkozio:
As you remember we discussed

> i18n: i18n.t(variable) - avoid, gettext tools expect i18n.t("text…")

And after discussion with @Akiyamka looks like it should be fixed because extractors may be used. You can find problematic places searching for ` i18n\.t\(\w`

Komzpa:
.pot file is usually put into `template` folder nearby the translations, not one level up - as project grows bigger there are sometimes file hierarchies under each of the folders.

propakov:
> As you remember we discussed
> 
> > i18n: i18n.t(variable) - avoid, gettext tools expect i18n.t("text…")
> 
> And after discussion with @Akiyamka looks like it should be fixed because extractors may be used. You can find problematic places searching for ` i18n\.t\(\w`

@vkozio Yep, we have next ticket about extracting translations, so we gonna fix it there
This one is about convertion to gettext and back.

propakov:
@Komzpa wow, nice catch! thanks!

vkozio (approved):

Komzpa (approved):
LGTM. I expect we will have another pass on this when actually will have exchange with the translators.

PR #257, opened 2022-11-11
fix(core, user_profile): 13179 - apply changes when user profile changes 

IlyaIzr:
Changes now applied when user logs in, out, and changes preferences

We have 3 possible choises that affect app so far: language, units (map scale) and OSM editor. They all reactivly updated on user change without reloading

Reloading is okay however in future if no-reloading implementation is dificuilt. I removed it just now as we don't seem to need it

### DIFF
--- a/src/features/user_profile/atoms/userProfile.ts
+++ b/src/features/user_profile/atoms/userProfile.ts
@@ -35,18 +35,14 @@ export const currentProfileAtom = createAtom(
           5,
         );
 
-        i18n.instance
-          .changeLanguage(user.language)
-          .then((r) => {
-            location.reload();
-          })
-          .catch((e) => {
-            console.error(e);
-          });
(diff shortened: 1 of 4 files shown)


src/features/user_profile/atoms/userProfile.ts
```diff
@@ -35,18 +35,14 @@ export const currentProfileAtom = createAtom(
...
-          })
-          .catch((e) => {
-            console.error(e);
-          });
+        dispatch(currentUserAtom.setUser(user));
       });
     });
 
     onChange('currentUserAtom', async (newUser, prevUser) => {
+      // reload to simply apply all the settings if implementation is difficult
```
propakov, line 43: is it forgotten to remove or we need it?

propakov (approved):

propakov:
tests are failing, pls fix and will be great. 👍

PR #259, opened 2022-11-15
fix(event_list): 13247 - Event list panel that takes all height for disasters panel

IlyaIzr:
I've added 100vh-height element to the end of the virtual list so that it would stretch up panel to maximum height (height limited with <SmartColumn />)
image

why not 100% height for panels container of `.smartColumnCocontent`? Because then we have overflow poblem - panels can be resize out of screen by height

also while trying 100% height for  `.smartColumnCocontent` i noticed a small bug at column context

### DIFF
--- a/src/features/events_list/components/EventsListPanel/EventsListPanel.tsx
+++ b/src/features/events_list/components/EventsListPanel/EventsListPanel.tsx
@@ -86,7 +86,6 @@ export function EventsListPanel({
         classes={panelClasses}
         isOpen={isOpen}
         modal={{ onModalClick: onPanelClose, showInModal: isMobile }}
-        minContentHeightPx={MIN_HEIGHT}
         resize={isMobile ? 'none' : 'vertical'}
         contentClassName={s.contentWrap}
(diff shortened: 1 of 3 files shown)


vkozio (approved):

Akiyamka:
Please add some explanation in code

src/features/events_list/components/EventsListPanel/EventsListPanel.tsx
```diff
@@ -101,23 +100,27 @@ export function EventsListPanel({
...
+                        alternativeActionControl={
+                          userModel?.hasFeature(AppFeature.EPISODES_TIMELINE) ? (
+                            <EpisodeTimelineToggle isActive={event.eventId === current} />
+                          ) : null
+                        }
+                      />
+                    )}
+                    ref={virtuoso}
+                  />
+                  {/* this element won't take visual or scroll space */}
```
Akiyamka: It would have been better to write what it is for and what it does, instead of describing what it does not do

PR #262, opened 2022-11-17
feat(core/routes): 13301 - implement subpages in mode

Akiyamka:
Modes was removed because it just routes in fact
Support of nested routes added
All routes conflagration now in one config file
image


src/core/auth/atoms/userWasLanded.ts
```diff
@@ -0,0 +1,27 @@
...
+import { userResourceAtom } from './userResource';
+
+export const userWasLandedAtom = createAtom(
+  {
+    userResourceAtom,
+  },
+  ({ get }, state: 'yes' | 'no' | 'pending' = 'pending') => {
+    const userModel = get('userResourceAtom');
+    const userLoaded = userModel.data && !userModel.loading;
+    if (userLoaded) {
```
vkozio, line 12: This specific landing redirect functionality is feature, not core thing, and I [planned to move in into Aboutmode feature](https://kontur.fibery.io/Tasks/Task/Refactor-About-mode-to-perform-redirect-inside-feature,-not-in-Routes-(decoupling)-13264) - it should be possible to disable it via feature flag.

Akiyamka: agree

src/core/router/components/Routes.tsx
```diff
@@ -0,0 +1,36 @@
...
+          <Suspense fallback={<h1>Fallback</h1>}>{r.view}</Suspense>
+        </UniversalRoute>
+      ))}
+    </>
+  );
+};
+
+export function Routes() {
+  const [availableRoutes] = useAtom(availableRoutesAtom);
+  if (availableRoutes === null) return <h1>Loading</h1>; // Maybe some loading screen needed?
```
vkozio: Those big <h1>Fallback</h1> and <h1>Loading</h1> doesn't look good during load, maybe leave them empty till we have better visual idea?

vkozio (changes requested):

src/views/Common.tsx
```diff
@@ -0,0 +1,69 @@
...
+import type { PropsWithChildren } from 'react';
+
+const { AppHeader } = lazily(() => import('@konturio/ui-kit'));
+const { NotificationToast } = lazily(() => import('~features/toasts'));
+const { PopupTooltip } = lazily(() => import('~features/tooltip'));
+const { SideBar } = lazily(() => import('~features/side_bar'));
+
+const DEFAULT_HEADER_TITLE = 'Disaster Ninja';
+
+export function Views() {
```
IlyaIzr: didn't you say if file has single export it should be == to filename?

src/views/Common.tsx
```diff
@@ -0,0 +1,69 @@
...
+      <OriginalLogo />
+
+      <Suspense fallback={null}>
+        {userModel?.hasFeature(AppFeature.HEADER) && (
+          <AppHeader
+            title={currentRoute?.customHeader ?? DEFAULT_HEADER_TITLE}
+            logo={VisibleLogo()}
+          />
+        )}
+      </Suspense>
```
IlyaIzr: we removed header from the app (delivered on prod already)

src/core/router/routes.tsx
```diff
@@ -0,0 +1,85 @@
...
+      slug: 'profile',
+      title: (
+        <UserStateToComponents
+          authorized={i18n.t('modes.profile')}
+          other={i18n.t('login.login_button')}
+        />
+      ),
+      icon: <User24 />,
+      view: <ProfilePage />,
+      requiredPermission: AppFeature.APP_LOGIN,
```
IlyaIzr: required feature name? It's wierd that _permission_ is a string

src/components/HeaderTitle/HeaderTitle.tsx
```diff
@@ -0,0 +1,22 @@
...
+    <Text type="short-l">
+      <div className={s.customAppTitle}>
+        <span className={s.clickable} onClick={toHomePage} title={i18n.t('to_main_page')}>
+          Disaster Ninja
+        </span>{' '}
+        {children}
+      </div>
+    </Text>
+  );
+}
```
IlyaIzr: we removed header

src/core/router/components/Routes.tsx
```diff
@@ -0,0 +1,36 @@
...
+          as={r.cached ? CacheRoute : Route}
+        >
+          <Suspense fallback={<h1>Fallback</h1>}>{r.view}</Suspense>
+        </UniversalRoute>
+      ))}
+    </>
+  );
+};
+
+export function Routes() {
```
IlyaIzr, line 48: can we call this component `<SelectedRoute />`?

src/views/About/About.tsx
```diff
@@ -1,19 +1,12 @@
 import { Trans } from 'react-i18next';
 import { i18n } from '~core/localization';
-import config from '~core/app_config';
-import history from '~core/history';
-import { APP_ROUTES } from '~core/app_config/appRoutes';
 import s from './About.module.css';
 
-function gotoMap() {
-  history.push(APP_ROUTES.map);
+export function AboutPage({ toHomePage }: { toHomePage: () => void }) {
```
propakov, line 7: do we really need AboutPage ? maybe lets remove it and rename AboutText -> AboutPage?

src/core/router/atoms/showAboutForNewUsers.ts
```diff
@@ -0,0 +1,22 @@
+import { createAtom } from '@reatom/core';
+import { userWasLandedAtom } from '~core/auth/atoms/userWasLanded';
+import history from '../history';
+import { getAbsoluteRoute } from '../routes';
+import { availableRoutesAtom } from './availableRoutes';
+
+export const showAboutForNewUsersAtom = createAtom(
```
propakov, line 7: it looks like this atom is never imported anywhere and never run

propakov: so that's why this redirection for new user not triggered

IlyaIzr:
also please check that reloading is working correctly: go to about page, reload page, try to stay on about page (i had redirection back to default path)

src/core/app_config/index.ts
```diff
@@ -52,7 +29,7 @@ export default (() => {
     layersByDefault: konturAppConfig.LAYERS_BY_DEFAULT,
     featuresByDefault: konturAppConfig.FEATURES_BY_DEFAULT,
     defaultFeed: konturAppConfig.DEFAULT_FEED,
-    defaultFeedObject: getDefaultFeedObject(konturAppConfig.DEFAULT_FEED),
+    defaultFeedObject: DEFAULT_FEED_OBJECT, // translation should occur later after i18n init, getDefaultFeedObject(konturAppConfig.DEFAULT_FEED),
```
propakov: found that it could be done this way
 ```
 i18n
  .use(LanguageDetector)
  .use(initReactI18next)
  .init(
    {
      fallbackLng: 'en',
      debug: false,
      defaultNS: 'common',
      interpolation: {
        escapeValue: false,
      },
      resources: {
        en,
        es,
        ar,
      },
    },
    (_, t) => {
      DEFAULT_FEED_OBJECT.name = t('logout');
    },
```
it's clumsy, but maybe for this unusual case it's ok ? just because I'm not sure if you want to await i18n before you get config
BTW do you have an idea of how to implement it later?

File: src/core/router/components/Routes.tsx (hunk 2 above)
Comment (propakov) [reply src/core/router/components/Routes.tsx]: yeah, it feels like Fallback tells nothing to user
some loading icons would suit better for both cases

File: src/features/side_bar/components/SideBar/SideBar.tsx [hunk 11]
```diff
@@ -63,84 +97,91 @@ export function SideBar() {
...
-          )}
+            )}
+          </div>
         </div>
 
         <div className={s.konturLogo}>
           <Logo compact={!isOpen} palette="grey" height={32} />
         </div>
       </ActionsBar>
+           )}
```
propakov: ooops, prettier lost 🤔

src/features/side_bar/components/SideBar/SideBar.tsx
```diff
@@ -1,28 +1,62 @@
...
+  return (route: AppRoute, currentRoute: AppRoute | null): boolean => {
+    if (route.hidden) return false;
+    if (!route.parentRoute) return true; // always show top level routes
+    if (currentRoute === null) return false; // hide nested routes if no selected routes
+    const isActive = route.slug === currentRoute.slug;
+    const haveActiveParentRoute = route.parentRoute
+      ? currentRoute?.slug === route.parentRoute
+      : false;
+    const neighbors = route.parentRoute ? Object.keys(routesTree[route.parentRoute]) : [];
+    const haveActiveNeighbor = neighbors.indexOf(currentRoute.slug) !== -1;
```
propakov:
```suggestion
    const haveActiveNeighbor = neighbors.includes(currentRoute.slug);
```

src/core/router/routes.tsx
```diff
@@ -0,0 +1,85 @@
...
+    {
+      slug: 'map',
+      title: i18n.t('modes.map'),
+      icon: <Map24 />,
+      view: <MapPage />,
+      cached: true,
+    },
+    {
+      slug: 'reports',
+      title: i18n.t('modes.reports'),
```
propakov, line 60: pls add en translation

src/features/side_bar/components/SideBar/SideBar.tsx
```diff
@@ -63,84 +97,91 @@ export function SideBar() {
...
                 </span>
               ) : null}
             </ActionsBarBTN>
+            </div>
           </div>
-        </div>
+          {availableRoutes.routes.map((route) => {
+            return checkRouteVisibility(route, currentRoute) ? (
+              <Link
+                key={nanoid(4)}
```
propakov: it looks like we can use key={route.slug}

src/core/router/routes.tsx
```diff
@@ -0,0 +1,85 @@
...
+      icon: <Alarm24 />,
+      view: <ReportsPage />,
+      customHeader: <HeaderTitle>{i18n.t('sidebar.reports')}</HeaderTitle>,
+      requiredPermission: AppFeature.REPORTS,
+    },
+    {
+      slug: ':reportId',
+      title: i18n.t('modes.report'),
```
propakov: modes.report is without translation

src/core/router/components/Routes.tsx
Akiyamka: Agree, i will add our spinner

src/views/Common.tsx
Akiyamka: nice catch!

src/views/Common.tsx
Akiyamka: got it, removed

src/core/router/routes.tsx
Akiyamka: why its wired?

Akiyamka: required permission renamed to requiredFeature

src/views/About/About.tsx
Akiyamka: agree

src/core/router/components/Routes.tsx
Akiyamka: Why we should name it like that? it's contain all routes

src/core/app_config/index.ts
Akiyamka: As I known @vkozio done it that way for solve cycled deps. 
I have an ideas how to solve it in general, it out of scope of this task

src/core/router/atoms/showAboutForNewUsers.ts
Akiyamka: fixed!

src/core/router/components/Routes.tsx
IlyaIzr: Yes...
It's just where i approximately thought routes config transformed to single returned component/view of the route before. (obviously it's not now)

It's something i'd expect to find in a components or just near them. Where is that happening?

IlyaIzr (changes requested):
Bug with reloading still reproduceable

src/core/router/components/Routes.tsx
Akiyamka: This logic inside react-router library

src/core/router/routes.tsx
```diff
@@ -0,0 +1,87 @@
...
+  Prefs24,
+  User24,
+  Reports16,
+} from '@konturio/default-icons';
+import { i18n } from '~core/localization';
+import { AppFeature } from '~core/auth/types';
+import { UserStateToComponents } from '~core/auth';
+import history from './history';
+import type { AppRouterConfig } from './types';
+// import app_config from '~core/app_config';
```
propakov: forgotten comment

src/core/app_config/postinit.ts
```diff
@@ -0,0 +1,15 @@
+import { i18n } from '~core/localization';
+
+export function getDefaultFeedObject(feed?: string) {
```
propakov: this file  is never imported
forgotten to remove ?

Akiyamka: hmm,  like we loose some functionality after rebase

propakov: ok, so need to be rechecked after returning

Akiyamka: Fixed. Not the best solution, but i add task about think where to write such logic
https://kontur.fibery.io/search/ZKZ06#Tasks/Task/Fix-cycled-deps-between-localization-service-and-config-13554

IlyaIzr (approved):

propakov:
@Akiyamka try to open any report 

Screenshot 2022-11-30 at 14 48 38

src/views/Map/Map.tsx
```diff
@@ -36,10 +38,14 @@ const { BivariatePanel } = lazily(() => import('~features/bivariate_manager/comp
...
 
-type MainViewProps = {
-  userModel: UserDataModel | null;
-};
-export function MainView({ userModel }: MainViewProps) {
+export function MapPage() {
+  const [{ data, loading }] = useAtom(userResourceAtom);
+  const userModel = data && !loading ? data : null;
+
+  useEffect(() => {
```
propakov, line 44: btw, why we init metrics not in CommonView where all common stuff located? We are sure that we won't need metrics on different screens?

Komzpa:
Reopening to trigger CI

src/views/Map/Map.tsx
vkozio: metrics should init at app start, not when we switch to map page from other page. metrics need to know current route and currently can process only map page

vkozio (approved):

propakov (approved):
👍

PR #269, opened 2022-11-27
fix(components, analytics_panel): 8798 - make links shorter

IlyaIzr:
Fixed bug appeared on links parsed that was only detected on 3+ number of links per parse.
Created component for short links
Links for analytics became shorter and 'www.' prefix is cutted off
Tests updated accordingly
Links before:
image
Links now on analytics panel
image

P.s. fix on `./Routes` is not meant to be part of this PR, it just won't load otherwise - so it's weird to push broken app. FIXME added and i'll remove it in case assigned developer would find better solution

### DIFF
--- a/src/Routes.tsx
+++ b/src/Routes.tsx
@@ -29,12 +29,19 @@ export function RoutedApp() {
   const [{ data, loading }] = useAtom(userResourceAtom);
   const userModel = data && !loading ? data : null;
 
+  // FIXME: this logic will be removed on PR #262 or by demand - to be discussed
+  // written as is it wouldn't even allow to start page
   useEffect(() => {
     if (userModel) {
       metricsInit();
-      return forceRun(urlStoreAtom);
+      // This one never runs initially
+      // return forceRun(urlStoreAtom);
     }
   }, [userModel]);
+
+  useEffect(() => {
+    forceRun(urlStoreAtom);
+  }, []);
 
   useEffect(() => {
     const isFirstTimeVisit = () =>
--- a/src/Routes.tsx
+++ b/src/Routes.tsx
@@ -29,19 +29,12 @@ export function RoutedApp() {
   const [{ data, loading }] = useAtom(userResourceAtom);
   const userModel = data && !loading ? data : null;
 
-  // FIXME: this logic will be removed on PR #262 or by demand - to be discussed
-  // written as is it wouldn't even allow to start page
   useEffect(() => {
     if (userModel) {
       metricsInit();
-      // This one never runs initially
-      // return forceRun(urlStoreAtom);
+      return forceRun(urlStoreAtom);
     }
   }, [userModel]);
-
-  useEffect(() => {
-    forceRun(urlStoreAtom);
-  }, []);
 
   useEffect(() => {
     const isFirstTimeVisit = () =>
--- a/src/components/LinkRenderer/LinkRenderer.module.css
+++ b/src/components/LinkRenderer/LinkRenderer.module.css
@@ -1,0 +1,35 @@
+.linkWidthWrap {
+  position: relative;
+  max-width: 192px;
+  & .linkOverflowWrap {
+    white-space: nowrap;
+    text-overflow: ellipsis;
+    overflow: hidden;
+  }
+
+  & .link {
+    white-space: nowrap;
+    text-overflow: ellipsis;
+    overflow: hidden;
+    width: 100%;
+    display: block;
+    text-decoration: none;
+    color: var(--accent-strong);
+    cursor: pointer;
+  }
+
+  & .link::after {
+    content: attr(data-truncate);
+    position: absolute;
+    left: 100%;
+    top: 0;
+    white-space: nowrap;
+  }
+
+  & .link:hover {
+    text-decoration: underline;
+    &::after {
+      text-decoration: underline;
+    }
+  }
+}
--- a/src/components/LinkRenderer/LinkRenderer.module.css
+++ b/src/components/LinkRenderer/LinkRenderer.module.css
@@ -1,6 +1,5 @@
 .linkWidthWrap {
   position: relative;
-  max-width: 192px;
   & .linkOverflowWrap {
     white-space: nowrap;
     text-overflow: ellipsis;
--- a/src/components/LinkRenderer/LinkRenderer.tsx
+++ b/src/components/LinkRenderer/LinkRenderer.tsx
@@ -1,19 +1,42 @@
 import { memo } from 'react';
+import s from './LinkRenderer.module.css';
 
-export const LinkRenderer = memo(function (props: any) {
+// types from react-markdown expected types
+type ReactMemoElementWrapProps = React.DetailedHTMLProps<
+  React.AnchorHTMLAttributes<HTMLAnchorElement>,
+  HTMLAnchorElement
+>;
+
+export const LinkRenderer = memo(function (props: ReactMemoElementWrapProps) {
   return (
-    <a
-      href={props.href}
-      target="_blank"
-      rel="noreferrer"
-      onClick={stopPropagation}
-    >
+    <a href={props.href} target="_blank" rel="noreferrer" onClick={stopPropagation}>
       {props.children}
     </a>
   );
 });
+
+export const LinkRendererShort = memo(function ({
+  children: linkArr,
+}: ReactMemoElementWrapProps) {
+  return (
+    <div className={s.linkWidthWrap}>
+      <div className={s.linkOverflowWrap}>
+        <a
+          className={s.link}
+          target="_blank"
+          rel="noreferrer"
+          data-truncate={linkArr?.[0].slice(-12)}
+        >
+          {linkArr?.[0]}
+        </a>
+      </div>
+    </div>
+  );
+});
+
+LinkRenderer.displayName = 'LinkRenderer';
+LinkRendererShort.displayName = 'LinkRendererShort';
+
 function stopPropagation(e) {
   e.stopPropagation();
 }
-
-LinkRenderer.displayName = 'LinkRenderer';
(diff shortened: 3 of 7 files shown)


src/components/LinkRenderer/LinkRenderer.tsx
```diff
@@ -1,19 +1,42 @@
 import { memo } from 'react';
+import s from './LinkRenderer.module.css';
 
-export const LinkRenderer = memo(function (props: any) {
+// types from react-markdown expected types
+type ReactMemoElementWrapProps = React.DetailedHTMLProps<
```
propakov: name of type feels a bit inconsistent, or maybe I just don't get it 🤷‍♂️ 
it's memo in the name but I see nothing about memo in type 🤔 
I would expect LinkRendererProps

src/Routes.tsx
```diff
@@ -29,13 +29,20 @@ export function RoutedApp() {
   const [{ data, loading }] = useAtom(userResourceAtom);
   const userModel = data && !loading ? data : null;
 
+  // FIXME: this logic will be removed on PR #262 or by demand - to be discussed
+  // written as is it wouldn't even allow to start page
```
propakov: this file should not be a part of this PR, as will be fixed by @vkozio 
need to not to forget to remove these changes

src/utils/markdown/parser.ts
```diff
@@ -14,29 +14,30 @@ export function parseLinksAsTags(text?: string): string {
...
-    const matchIndex = matchEntity.index!;
+    const matchIndex = matchEntity.index ?? 0;
     const matchLength = match.length;
     // skip links in propper markdown format
     if (match.startsWith('](') || match.indexOf('[http') > -1) return;
     // get full link to work with it
     const linkStartIndex = match.indexOf('http');
     const fullLink = match.substring(linkStartIndex);
     const beforeLink = match.substring(0, linkStartIndex);
+    const minimalisticDomain = domain.replace('www.', '');
```
propakov: maybe domainNoW3/ domainWithoutWWW name would be more concrete

src/components/LinkRenderer/LinkRenderer.tsx
```diff
@@ -1,19 +1,42 @@
...
+  children: linkArr,
+}: ReactMemoElementWrapProps) {
+  return (
+    <div className={s.linkWidthWrap}>
+      <div className={s.linkOverflowWrap}>
+        <a
+          className={s.link}
+          target="_blank"
+          rel="noreferrer"
+          data-truncate={linkArr?.[0].slice(-12)}
```
propakov: why -12? I would like to see any comments there

src/components/LinkRenderer/LinkRenderer.module.css
```diff
@@ -0,0 +1,35 @@
+.linkWidthWrap {
+  position: relative;
+  max-width: 192px;
```
vkozio: Do we have some constant/var for this maybe?

src/utils/markdown/parser.ts
```diff
@@ -14,29 +14,30 @@ export function parseLinksAsTags(text?: string): string {
...
+      matchIndex + offset,
+      matchLength,
+      `${beforeLink}[${minimalisticDomain}${path ?? ''}](${fullLink})`,
     );
 
-    pointer +=
-      `[${domain}${path ?? ''}](${fullLink})`.length -
-      (matchLength - beforeLink.length);
+    offset +=
+      `${beforeLink}[${minimalisticDomain}${path ?? ''}](${fullLink})`.length -
```
propakov: this complex string expression is repeated twice, let's move it to variable
```
      `${beforeLink}[${minimalisticDomain}${path ?? ''}](${fullLink})`

File: src/components/LinkRenderer/LinkRenderer.tsx (hunk 1 above)
Comment (IlyaIzr) [reply src/components/LinkRenderer/LinkRenderer.tsx]: hehe fair enough, the naming means those are types that library 'react-memo' expects
i'll fix it

File: src/components/LinkRenderer/LinkRenderer.tsx (hunk 4 above)
Comment (IlyaIzr) [reply src/components/LinkRenderer/LinkRenderer.tsx]: okay!
it was suggested by the task i'm implementing

File: src/components/LinkRenderer/LinkRenderer.module.css (hunk 5 above)
Comment (IlyaIzr) [reply src/components/LinkRenderer/LinkRenderer.module.css]: hm,  not really, it was specific value asked in the task for link width

File: src/components/LinkRenderer/LinkRenderer.tsx (hunk 1 above)
Comment (IlyaIzr) [reply src/components/LinkRenderer/LinkRenderer.tsx]: whops, it was supposed to be react-markdown... sorry, thanks!

File: src/components/LinkRenderer/LinkRenderer.module.css (hunk 5 above)
Comment (propakov) [reply src/components/LinkRenderer/LinkRenderer.module.css]: it would be good to have a prop "max-width" for LinkRendererShort (with 192 default value)
like how short link should be
at least it would be configurable

**Comment by propakov** [comment]
I've found next warning in console:
`react_devtools_backend.js:4026 Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.`
i believe it should be fixed

File: src/components/LinkRenderer/LinkRenderer.module.css (hunk 5 above)
Comment (propakov) [inline src/components/LinkRenderer/LinkRenderer.module.css]: as you moved it to the props - this style is redundant then

File: src/components/LinkRenderer/LinkRenderer.tsx [hunk 7]
```diff
@@ -1,17 +1,44 @@
...
-      onClick={stopPropagation}
-    >
+    <a href={props.href} target="_blank" rel="noreferrer" onClick={stopPropagation}>
       {props.children}
     </a>
   );
 });
+
+// 192 and 12 are the values proposed by the task #8798
+export function getShortLinkRenderer(
```
propakov: oooh, it looks so verbose 😞 

I suggest next: 
- as for now we are ok with default options we use it like it was 
```
 <ReactMarkdown
                components={{ a: ShortLinkRenderer }}
 ```
 in case we need to pass options 
 ```
 components={{ a: (props) => <ShortLinkRenderer maxWidth={100} {...props} /> }}
 ```
 then we can use component without wrapper 
 ```
 type ShortLinkRenderer = ElementWrapProps & {
  maxWidth?: number;
  truncateAmount?: number;
};

export function ShortLinkRenderer({
  children: linkArr,
  maxWidth = 192,
  truncateAmount = 12,
}: ShortLinkRenderer) {
 ```

File: src/components/LinkRenderer/LinkRenderer.tsx [hunk 8]
```diff
@@ -1,17 +1,44 @@
...
-      rel="noreferrer"
-      onClick={stopPropagation}
-    >
+    <a href={props.href} target="_blank" rel="noreferrer" onClick={stopPropagation}>
       {props.children}
     </a>
   );
 });
+
+// 192 and 12 are the values proposed by the task #8798
```
propakov, line 18: We have task number in the commit message, so this comment seems not that useful.
I meant to point out that we slice it from the end of the string.

src/components/LinkRenderer/LinkRenderer.tsx
```diff
@@ -1,17 +1,44 @@
...
+    return (
+      <div className={s.linkWidthWrap} style={style}>
+        <div className={s.linkOverflowWrap}>
+          <a
+            className={s.link}
+            target="_blank"
+            rel="noreferrer"
+            data-truncate={linkArr?.[0].slice(-truncateAmount)}
+          >
+            {linkArr?.[0]}
```
propakov: maybe let's put to the variables with names, so no need comments then 🤔 
const link = linkArr?.[0];
const truncatedLinkEnd = link.slice(-truncateAmount);

propakov (changes requested):

src/components/LinkRenderer/LinkRenderer.tsx
IlyaIzr: can be handy if someone don't have git-lense extension

propakov (commented):
Code LGTM.  🔥  Just warning needs to be fixed.

src/components/LinkRenderer/LinkRenderer.tsx
```diff
@@ -1,17 +1,45 @@
...
 });
+
+// 192 and 12 are the values proposed by the task #8798
+export function ShortLinkRenderer({
+  children: linksArr,
+  maxWidth = 192,
+  truncateAmount = 12,
+}: ElementWrapProps & { maxWidth?: number; truncateAmount?: number }) {
+  const style = { maxWidth: maxWidth || 'unset' };
+  const passedLink: string | undefined = linksArr?.[0];
```
Akiyamka, line 29: It's not clear for me why we process only one link from array?

Akiyamka (approved):
Done well!

Akiyamka (changes requested):
Panel should stay common. Create root components that known about invariants and manage them. In simple worlds - you need a component that manage what state of panel should be visible right now.

propakov:
click on the link isn't working. 
it doesn't feel like expected behaviour
Screenshot 2022-11-30 at 09 20 25

propakov:
Also as link can't be selected like a normal text, lets add 
` user-select: none;  `
 to prevent it?

IlyaIzr:
> click on the link isn't working. it doesn't feel like expected behaviour Screenshot 2022-11-30 at 09 20 25

thanks you! my bad, fixed it. Text selecting is now usual to link also

Akiyamka (approved):

propakov (approved):
Yeah, link open works! 
I've meant, try to select whole link. It's interrupted before "..."
It feels a bit broken from ux, I would disable text selection for such links. But it's up to you.
image

IlyaIzr:
> Yeah, link open works! I've meant, try to select whole link. It's interrupted before "..." It feels a bit broken from ux, I would disable text selection for such links. But it's up to you. image


I'd like to keep it: while it's not giving a selection of the whole link, user can found it out easily; however he can still select part of the link for example to paste it into url bar or browser history and work with results

PR #270, opened 2022-11-28
chore(event_list): 12482 - short state for event_list panel

IlyaIzr:
How it looks:
dn_third_panel_state
Design solution:
create `ShortPanel` component. Make short content rely on `currentEventResourceAtom`
Question to reviewer:
is that a good solution? Possible options - add `eventListResourceAtom` (then what's the priority to pick event), rely solely on `eventListResourceAtom` (i think it's wrong because not all events are in event list - old events has links and can be reached through `currentEventResourceAtom`)


IlyaIzr:
i will resolve merge conflict after fix of main branch expected from @vkozio 
it's only for translations anyway

src/core/localization/translations/en/common-messages.json
```diff
@@ -76,7 +76,9 @@
         "settled_area.tooltip": "Settled Area",
         "loss.tooltip": "Estimated loss"
       },
-      "no_event_in_feed": "Event not found in the current event feed"
+      "no_event_in_feed": "Event not found in the current event feed",
```
propakov: I see that you branched out not from the latest main
You need it for lint-staged step to update .pot file with the new keys

src/features/events_list/components/ShortPanel/ShortPanel.tsx
```diff
@@ -0,0 +1,59 @@
...
+          <Text type="short-m">{i18n.t('event_list.chose_disaster')}</Text>
+        </Button>
+      </div>
+    </div>
+  );
+
+  return <div className={s.shortPanel}>{panelContent}</div>;
+}
+
+function noop() {
```
propakov: suggest just use lodash noop

propakov: or it's need to be in utils, just not to have 10 noop functions around the project

src/core/localization/translations/en/common-messages.json
propakov: now pls just run` npm run i18n:export` 
it will update .pot file

IlyaIzr: did it!

src/features/events_list/components/EventsListPanel/EventsListPanel.module.css
```diff
@@ -1,8 +1,9 @@
 .panelContainer {
   display: flex;
   flex-flow: column nowrap;
-  &.open {
-    height: 100%;
+  &.short .contentWrap {
```
propakov: it looks like you don't need this override and .short class at all 
you can use Panel prop
        minContentHeightPx={isShortStateOpen ? 'min-content' : undefined} 
 it just get only number, maybe it makes sense to minContentHeightPx -> minContentHeight and get string too 🤔

propakov:
Regarding architecture:
I see that currentEventResourceAtom returns event with geometry, but event list use just events. 
Do you need event with geometry for short panel?
Because we have currentEventAtom that is updated from different places like autoSelectEvent. And for loading you can reuse loading prop from EventsListPanel, because whole list is loaded even in short panel mode.

src/core/types/index.ts
```diff
@@ -73,11 +73,17 @@ export interface AdvancedAnalyticsDataValues {
   quality: number;
 }
 export type EventWithGeometry = {
+  location: string;
   eventId: string;
   eventName: string;
   externalUrls: string[];
   severity: Severity;
   geojson: GeoJSON.GeoJSON;
+
```
Akiyamka: extra empty line

IlyaIzr:
to @propakov:
We sure don't need geometry in analytics
Problem with current event atom is that it only returns event id, but i need more event info. Possible solution is to 
1. Get current event id
2. Try find it in event list
3. If no result - get data from `currentEventResourceAtom`

Is that enough?

src/features/events_list/components/ShortPanel/ShortPanel.tsx
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  const panelContent = statesToComponents({
+    loading: <LoadingSpinner message={i18n.t('loading_events')} />,
+    error: (errorMessage) => <ErrorMessage message={errorMessage} />,
+    ready: (currentEvent) => (
+      <EventCard
+        event={currentEvent}
+        isActive={true}
+        onClick={noop}
```
Akiyamka: why not make onClick optional instead ?

src/features/events_list/components/ShortPanel/ShortPanel.module.css
```diff
@@ -0,0 +1,22 @@
+.shortPanel {
+  width: 100%;
+  height: min-content;
+}
+
+.noDisasters {
+  text-align: center;
+  height: 75px;
```
Akiyamka: the size of the block with text must not be fixed

propakov:
> to @propakov: We sure don't need geometry in analytics Problem with current event atom is that it only returns event id, but i need more event info. Possible solution is to
> 
> 1. Get current event id
> 2. Try find it in event list
> 3. If no result - get data from `currentEventResourceAtom`
> 
> Is that enough?

@IlyaIzr  Looks good, but it would be weird if you'll have event loaded in resource in shortPanel and don't have it when switch back to the list, isn't it? Or it's what we need?

Akiyamka (commented):
It would be better to nest short and long states in one parent order component that able to manage panel states

src/features/events_list/components/EventsListPanel/EventsListPanel.module.css
IlyaIzr: unfortunately i need also to reset inline property of 'height' for content container with this ugly `!important`

src/features/events_list/components/EventsPanel/EventsPanel.module.css
```diff
@@ -1,8 +1,9 @@
 .panelContainer {
   display: flex;
   flex-flow: column nowrap;
-  &.open {
-    height: 100%;
+  &.short .contentWrap {
```
propakov, line 4: forgotten to remove

src/features/events_list/components/FullState/FullState.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const statesToComponents = createStateMap({
+    error,
+    loading,
+    data: eventsList,
+  });
+
+  // Virtual event list rendering effect
+  useEffect(() => {
+    // type any is used because virtuoso types doesn't have scrollToIndex method, but it's described in docs https://virtuoso.dev/scroll-to-index
+    const ref: any = virtuoso.current;
```
propakov: suggest to remove any and comment and use 
```
const virtuoso = useRef<VirtuosoHandle>(null);
```
like here https://virtuoso.dev/virtuoso-api-reference/

src/features/events_list/components/EventsPanel/EventsPanel.tsx
```diff
@@ -0,0 +1,87 @@
...
+  }, [setPanelState]);
+
+  const onPanelClose = useCallback(() => {
+    setPanelState('closed');
+  }, [setPanelState]);
+
+  useAutoCollapsePanel(isOpen, onPanelClose);
+
+  return (
+    <div className={clsx(s.panelContainer, s[panelState])}>
```
propakov: do you really need s[panelState] here? 
I can't see even 1 case you use it in EventsPanel.module.css

src/features/events_list/components/EventsPanel/EventsPanel.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { AppFeature } from '~core/auth/types';
+import { IS_MOBILE_QUERY, useMediaQuery } from '~utils/hooks/useMediaQuery';
+import { useAutoCollapsePanel } from '~utils/hooks/useAutoCollapsePanel';
+import { panelClasses } from '~components/Panel';
+import { useHeightResizer } from '~utils/hooks/useResizer';
+import { MIN_HEIGHT } from '../../constants';
+import { ShortState } from '../ShortState/ShortState';
+import { FullState } from '../FullState/FullState';
+import s from './EventsPanel.module.css';
+import type { ShortStateListenersType } from '@konturio/ui-kit/tslib/Panel';
```
propakov: it should be `from '@konturio/ui-kit'`;
it's not exported in index.ts in ui repository

src/features/events_list/components/EventsListPanel/EventsListPanel.module.css
propakov: i've just removed this block and can't find short state broken or something
could you please clarify how we need it?

src/features/events_list/components/ShortState/ShortState.tsx
```diff
@@ -0,0 +1,90 @@
...
+  // Try get event from event list
+  useEffect(() => {
+    if (!event && currentEventId && eventsList.data) {
+      const eventFromList = eventsList.data.find((ev) => ev.eventId === currentEventId);
+      if (eventFromList) setEvent(eventFromList);
+    } else if (event && currentEventId === null) setEvent(null);
+  }, [eventsList, currentEventId, setEvent]);
+
+  // Try get event from currentEventResourceAtom
+  useEffect(() => {
```
propakov, line 54: in that case when you switch back to FullState you won't see this event in the list.
is it ok?

src/features/events_list/components/EventsListPanel/EventsListPanel.module.css
IlyaIzr: without it short state would be higher than its content
did you pull my latest changes?

src/features/events_list/components/EventsPanel/EventsPanel.tsx
IlyaIzr: will be fixed with component update

src/features/events_list/components/EventsPanel/EventsPanel.tsx
IlyaIzr: hmm, smth weird with mine or yours .css file

src/features/events_list/components/FullState/FullState.tsx
IlyaIzr: wow, they've updated types, cool!

src/features/events_list/components/EventsPanel/EventsPanel.tsx
IlyaIzr: image
image

src/features/events_list/components/ShortState/ShortState.tsx
IlyaIzr: Good question, answer is - yes, it's okay, but it will be developed soon https://kontur.fibery.io/Tasks/Task/Show-outdated-event-via-link-12519

src/features/events_list/components/EventsListPanel/EventsListPanel.module.css
propakov: yeah, found it. 
maybe it makes sense to do in throught panel props? 
we already have minContentHeight, so maybe it's a good reason to introduce contentHeight prop? so no need overrides in css then
what do you think @Akiyamka ? does it worth it, or we'll go with current approach?

src/features/events_list/components/EventsPanel/EventsPanel.tsx
propakov: yeah, I've got it. was sure we don't need it, wrong question.

src/features/events_list/components/ShortState/ShortState.tsx
propakov: ok, good to know, thank you!

Komzpa:
Reopening to trigger CI

src/features/events_list/components/EventsListPanel/EventsListPanel.module.css
IlyaIzr: I'll try to not forget to add it, sounds reasonable

src/features/events_list/components/EventsPanel/EventsPanel.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const isOpen = panelState !== 'closed';
+  const isMobile = useMediaQuery(IS_MOBILE_QUERY);
+  const [{ data: userModel }] = useAtom(userResourceAtom);
+  const handleRefChange = useHeightResizer(
+    (isOpen) => !isOpen && setPanelState('closed'),
+    isOpen,
+    MIN_HEIGHT,
+  );
+  const openFullState = () => setPanelState('full');
```
propakov: openFullState and onPanelIconClick do the same thing 
feels like 1 of them is redundant

propakov (approved):
good job! 👍  just 1 little moment to check

PR #272, opened 2022-11-28
fix(core): 13529 400 errors are returned from /active/api/rum/metrics if user is logged in

vkozio:
https://kontur.fibery.io/Tasks/Task/400-errors-are-returned-from-active-api-rum-metrics-if-user-is-logged-in-13529

### DIFF
--- a/src/core/metrics/app-metrics.ts
+++ b/src/core/metrics/app-metrics.ts
@@ -9,6 +9,7 @@ import {
   EVENT_MAP_IDLE,
 } from './constants';
 import { Sequence } from './sequence';
+import type { Unsubscribe } from '@reatom/core';
 import type { MetricsReportTemplate, MetricsEvent } from './types';
 
 const APP_METRICS_ENDPOINT = appConfig.apiGateway + '/rum/metrics';
@@ -50,8 +51,8 @@ export class AppMetrics {
   private eventLog: string[] = [];
   private settings = new SessionSettings<'KONTUR_SQ_ALERT' | 'KONTUR_SQ_LOG'>();
   reportTemplate: MetricsReportTemplate = METRICS_REPORT_TEMPLATE;
-  listener: void;
   private mode = '';
+  private unsubscribeCurrentModeAtom: Unsubscribe;
 
   static getInstance() {
     if (this._instance) {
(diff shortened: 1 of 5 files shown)


src/core/metrics/app-metrics.ts
```diff
@@ -127,6 +128,11 @@ export class AppMetrics {
   watchList = METRICS_WATCH_LIST;
 
   watch(name: string) {
+    // TODO: implement watchlists for other modes if necessary
+    // currently only map mode supported
+    const mode = currentModeAtom.getState();
```
IlyaIzr: `.getState()` recalculates atom state and subscribers state (correct me if i'm wrong @Akiyamka ). It used either in debug or when fired once\rarely. Is that the case?

src/core/metrics/init.ts
```diff
@@ -1,10 +1,9 @@
 import { currentApplicationAtom, currentUserAtom } from '~core/shared_state';
 import { appMetrics } from '~core/metrics';
-
 // FIXME: refactor like metricsInit(appId,user) without using atoms after app init implementation
 export function metricsInit() {
   const user = currentUserAtom.getState();
   const appId = currentApplicationAtom.getState();
   // metrics starts here because now we have true user id and app id
-  appMetrics.init(appId ?? '', user?.username ?? null);
+  appMetrics.init(appId ?? '', user?.email ?? null);
```
IlyaIzr, line 8: here's .init method
`init(appId: string, userId: string | null)`
Something is wrong :)

vkozio: we agreed with BE that userId is email

src/core/metrics/app-metrics.ts
vkozio: it 
> Read (or recalculate) state

https://github.com/artalar/reatom/blob/v2/packages/core/src/types.ts#L93

no subscribing

src/core/metrics/init.ts
IlyaIzr: it should be changed on init function then (and in according class)
any other person reading this code would think it's a bug

vkozio: It will be changed after app init refactor and implementing currentUser as SSoT of user data

src/core/metrics/app-metrics.ts
IlyaIzr: anyway it runs dozens of times (50+)  on app initiating, it's not okay
can you solve this different way?

vkozio: solved via subscribe

src/core/metrics/app-metrics.ts
```diff
@@ -71,11 +73,13 @@ export class AppMetrics {
     this.listener = globalThis.addEventListener(METRICS_EVENT, ((e: MetricsEvent) => {
       this.processEvent(e.detail.name, e.detail.payload);
     }) as EventListener);
+    currentModeAtom.subscribe((mode) => (this.mode = mode));
```
IlyaIzr: Cool! One more thing - whenever we have subscribe, we also need cleanup\unsubscribe mechanism

vkozio: cleanup implemented

src/core/metrics/app-metrics.ts
```diff
@@ -68,19 +71,29 @@ export class AppMetrics {
...
       console.info('appMetrics.init', this.reportTemplate);
     }
   }
 
+  // remove listeners and unsubscribe from atoms
+  cleanup() {
+    globalThis.removeEventListener(METRICS_EVENT, this.listener as EventListener);
+    this.unsubscribeCurrentModeAtom();
+  }
+
```
IlyaIzr, line 96: but is it called anywhere?

IlyaIzr (approved):

PR #274, opened 2022-11-30
chore(widgets, layers_panel, legend_panel): 12481 - merge layers and legend panels

IlyaIzr:
Working example of a widget that combines 2 features
Features don't need to communicate between each other so it's pretty simple
dn_layers_and_legends


IlyaIzr:
closing because change of base branch increased the filelist to review (and that list is wrong)

PR #275, opened 2022-12-01
feat(bivariate-legend): 12843 - change legend labels formatting

propakov:
https://kontur.fibery.io/Tasks/Task/Display-units-on-FE-DN-12843
image
image


src/components/BivariateLegend/BivariateLegend.tsx
```diff
@@ -115,3 +114,29 @@ export function BivariateLegend({
...
+    return `${nominator.label} (${nominator.unit.shortName})`;
+  }
+  // we have nominator unit, but don't have denominator unit - show only nominator unit
+  if (isNilOrOther(denominator.unit.id)) {
+    return `${nominator.label} to ${denominator.label} (${nominator.unit.shortName})`;
+  }
+  // here we process cases where we have both units
+  if (denominator.name === 'area_km2') {
+    return `${nominator.label} (${nominator.unit.shortName}/${denominator.unit.shortName})`;
+  }
```
IlyaIzr: great thing that you've used comments (bivariate stuff is hard by itself)
but you keep repeating "we have\don't have nominator unit"
maybe there's a way to tell that via code naming

src/components/BivariateLegend/BivariateLegend.tsx
```diff
@@ -115,3 +114,29 @@ export function BivariateLegend({
...
+
+  return `${nominator.label} to ${denominator.label} (${nominator.unit.shortName}/${denominator.unit.shortName})`;
+};
+
+const isNilOrOther = (unitId: string | null): boolean => !unitId || unitId === 'other';
```
IlyaIzr: since we're implementing new test coverage policy that function looks like a candidate for a simple test

src/utils/bivariate/types/stat.types.ts
```diff
@@ -17,12 +12,14 @@ export type Step = {
...
 
-/* Divisor and denominator pair */
+/* Nominator and denominator pair */
+type Unit = { id: string | null; longName: string | null; shortName: string | null };
 export type Quotient = [string, string]; // this field will be removed in next tickets, only Quotients will stay
 export type Quotients = {
   name: string;
   label: string;
   direction: Direction;
+  unit: Unit;
```
IlyaIzr, line 22: are we certain it can't be `undefined`? as you wrote "we don't have nominator unit" before, even though it's about `unit.id` i should still check

src/components/BivariateLegend/BivariateLegend.tsx
```diff
@@ -115,3 +114,29 @@ export function BivariateLegend({
     </div>
   );
 }
+
+const formatBivariateAxisLabel = (quotients: Axis['quotients']): string => {
+  if (!quotients) return '';
+  const [nominator, denominator] = quotients;
```
Komzpa: It is numerator not nominator.

src/components/BivariateLegend/BivariateLegend.tsx
propakov: ok, will add!

src/utils/bivariate/types/stat.types.ts
propakov: yeah, it always exists, checked this point

src/components/BivariateLegend/BivariateLegend.tsx
propakov: oh, my bad 🤦 thanks for this catch!

src/components/BivariateLegend/BivariateLegend.tsx
propakov: I have no idea how. If you have idea - please suggest.
Every 'if' just cuts down cases before 'if', that's why I've decided to describe what's covered in every case for easier navigation and understanding.

src/components/BivariateLegend/BivariateLegend.tsx
propakov: done

src/components/BivariateLegend/BivariateLegend.tsx
propakov: i've made them shorter, hope it will be ok

src/utils/bivariate/bivariateLegendUtils.ts
```diff
@@ -74,13 +49,9 @@ export function createBivariateMeta(
   yDenominator: string,
   stats: Stat,
 ): LayerMeta {
-  const xAxisNominatorIndicator = stats.indicators.find(
-    (ind) => ind.name === xNumerator,
-  );
+  const xAxisNominatorIndicator = stats.indicators.find((ind) => ind.name === xNumerator);
```
Komzpa: Nominator -> Numerator? maybe add a linter rule that "Nominator" is not expected anywhere in codebase?

src/utils/bivariate/labelFormatters.test.ts
```diff
@@ -0,0 +1,149 @@
+import { expect, describe, it } from 'vitest';
```
Komzpa, line 1: If you rebase this on latest main we can have a look how good is the coverage now :)

src/utils/bivariate/types/stat.types.ts
```diff
@@ -17,12 +12,14 @@ export type Step = {
   value: number;
 };
 
-/* Divisor and denominator pair */
+/* Nominator and denominator pair */
```
Komzpa: also divisor?

@migrek maybe we need to have a pass on the docs on this, seems that we have confusion on the codebase?

src/utils/bivariate/bivariateLegendUtils.ts
propakov: ah, that's the place where I've got this! 
yeah, lemme search it through the codebase

Komzpa (changes requested):
Naming of числитель is a concern in the codebase, can be fixed not now but need a ticket then.

A rebase would help have a look at test runs and improvement in coverage :)

src/utils/bivariate/types/stat.types.ts
migrek: Currently, we use "numerator" and "denominator" terms, but we can use "divisor" together with "dividend" as an alternative (doc with terms https://kontur.fibery.io/Tasks/project/Bivariate-Manager-441).
So, the 1st option looks better to me as terms are more distinguishable.

src/utils/bivariate/bivariateLegendUtils.ts
propakov: Screenshot 2022-12-02 at 16 33 43

src/utils/bivariate/labelFormatters.ts
```diff
@@ -7,3 +8,30 @@ export const convertDirectionsArrayToLabel = (directions: string[][]) => {
...
+  // cases for both units
+  if (denominator.name === 'area_km2') {
+    return `${numerator.label} (${numerator.unit.shortName}/${denominator.unit.shortName})`;
+  }
+
+  return `${numerator.label} to ${denominator.label} (${numerator.unit.shortName}/${denominator.unit.shortName})`;
+};
+
+export const isNilOrOther = (unitId: string | null): boolean =>
+  !unitId || unitId === 'other';
```
IlyaIzr: May be smth like this
I'm changing negative `isNilOrOther ` to avoid nesting negative statements
```suggestion

  // don't show units if there are none
  if (!hasUnits(nominator.unit.id)) {
    return `${nominator.label} to ${denominator.label}`;
  }
  
  if (denominator.name === 'one') {
    return `${nominator.label} (${nominator.unit.shortName})`;
  }
  // if no denominator unit - show only nominator unit
  if (!hasUnits(denominator.unit.id)) {
    return `${nominator.label} to ${denominator.label} (${nominator.unit.shortName})`;
  }
  // here we process cases where we have both units
  if (denominator.name === 'area_km2') {
    return `${nominator.label} (${nominator.unit.shortName}/${denominator.unit.shortName})`;
  }

  return `${nominator.label} to ${denominator.label} (${nominator.unit.shortName}/${denominator.unit.shortName})`;
};

const hasUnits = (unitId: string | null): boolean => Boolean(unitId && unitId !== 'other');
```

IlyaIzr (approved):
Since @Komzpa requested changes considering naming, I can approve the codes logical part. I've suggested what I think a better readable code piece, so I'd expect you to improve readability based on that. Approving all the rest!

src/utils/bivariate/labelFormatters.ts
propakov: hm, as far as I remember from clean code it's kinda a bad smell 
https://github.com/ryanmcdermott/clean-code-javascript#avoid-negative-conditionals
https://www.obsidiansecurity.com/blog/why-to-stop-writing-negative-code/

propakov:
@Komzpa fixed all "nominator" cases across the codebase

src/utils/bivariate/labelFormatters.test.ts
Komzpa: https://app.codecov.io/gh/konturio/disaster-ninja-fe/blob/12843-display-units-on-fe-dn/src/utils/bivariate/labelFormatters.ts it does not seem that this test is running as part of the CI suite, as it is not present in the coverage report or the lines of labelFormatters.ts are not shown as run. Can you plug it in so it actually runs as part of full test run?

Komzpa (changes requested):
It looks like the test in the PR is not running as part of the test suite.

src/core/localization/gettext/ar/common.po
```diff
@@ -403,10 +403,6 @@ msgstr "Stddev"
 msgid "Median"
 msgstr "الوسيط"
 
-#: advanced_analytics_data_list##filter_nominator
-msgid "Filter Nominator"
```
Komzpa, line 407: With normal gettext flows this should get a fuzzy flag or get commented out as "obsolete" (commented with `#~` prefix) so that it can resurface later. This can be achieved if you run msgmerge between the newly converted file and the one in repos and not just plainly replace.

Komzpa: https://stackoverflow.com/questions/10435926/how-to-automatically-remove-unused-gettext-strings

src/utils/bivariate/labelFormatters.ts
IlyaIzr: Exactly, and `isNillOrOther` is a negative code. And in case you'll ever go with `!isNillOrOther` it would be even worse. 
`if (!condition)` is negative too, but much more convenient for developers to read because of `!` to reverse statement

propakov: Ah, I thought it's more about usage with `!` in if statements, that it adds complexity.

propakov: thanks for explanation! 👍

src/core/localization/gettext/ar/common.po
propakov: idk why but can't make this command working in short time. 
I've marked by hands for now. it feels like we need to add such command in our chain.

src/utils/bivariate/labelFormatters.test.ts
propakov: yep, it was my bad. fixed now

Komzpa (approved):
LGTM

PR #276, opened 2022-12-05
chore(widgets, layers_panel, legend_panel): 12481 - merge layers and legend panels

IlyaIzr:
2 panels are now exist as panel content and combined by widget that is a panel
image


src/components/Panel/styles.module.css
```diff
@@ -1,7 +1,8 @@
 .header {
-  display: flex;
-  align-items: center;
   min-height: 40px;
+  & > div {
```
Akiyamka: better to add class name if possible.

src/widgets/LayersAndLegends/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+    <LayersAndLegends
+      legendMinHeight={legendPanelInterface.minHeight}
+      legendPanelContent={legendFeatureIsOn && legendPanelInterface.content}
+      legendIcon={legendFeatureIsOn && legendPanelInterface.panelIcon}
+      legendHeader={legendPanelInterface.header}
+      layersPanelContent={layersFeatureIsOn && layersPanelInterface.content}
+      layersIcon={layersFeatureIsOn && layersPanelInterface.panelIcon}
+      layersMinHeight={layersPanelInterface.minHeight}
+      layersHeader={layersPanelInterface.header}
+    />
```
Akiyamka: this is unnecessary wrapper if you pas interface directly into component

src/widgets/LayersAndLegends/components/LayersAndLegends.tsx
```diff
@@ -0,0 +1,99 @@
...
+    header?: string | JSX.Element;
+    minHeight?: number;
+  };
+  layersProps?: {
+    content?: JSX.Element | null | false;
+    panelIcon?: JSX.Element | false;
+    header?: string | JSX.Element;
+    minHeight?: number;
+  };
+};
```
vkozio:
```
{
legendProps?: PanelFeatureInterface;
layersProps?: PanelFeatureInterface;
}
```

You already implemented PanelFeatureInterface to use in such places

src/widgets/LayersAndLegends/components/LayersAndLegends.tsx
```diff
@@ -0,0 +1,99 @@
...
+    minHeight: layersMinHeight,
+    header: layersHeader,
+    panelIcon: layersIcon,
+  } = layersProps || {};
+  const {
+    content: legendPanelContent,
+    minHeight: legendMinHeight,
+    header: legendHeader,
+    panelIcon: legendIcon,
+  } = legendProps || {};
```
vkozio: same here, PanelFeatureInterface. Otherwise you'll need to fix such types each time PanelFeatureInterface changes

vkozio (changes requested):

src/widgets/LayersAndLegends/index.tsx
```diff
@@ -0,0 +1,22 @@
+import { LayersAndLegends } from './components/LayersAndLegends';
```
propakov: it differs from our convention we have. 
let's see how views organized, we import ~views/BivariateManager/BivariateManager
it seems like  you need to LayersAndLegends/index.tsx -> LayersAndLegendsWidget/LayersAndLegendsWidget.tsx

cosmos.vite.config.ts
```diff
@@ -57,6 +57,7 @@ const config: UserConfigFn = async () => {
         '~appModule': relative('./src/redux-modules/appModule'),
         '~core': relative('./src/core'),
         '~features': relative('./src/features'),
+        '~widgets': relative('./src/widgets'),
```
propakov, line 60: 🙏

src/features/layers_panel/index.tsx
```diff
@@ -1,5 +1,14 @@
...
+import { lazily } from 'react-lazily';
+import { i18n } from '~core/localization';
+import { MIN_HEIGHT } from './constants';
+import type { PanelFeatureInterface } from 'types/featuresTypes';
 
-export function MapLayersList() {
-  return <MapLayerPanel />;
-}
+const { PanelContent } = lazily(() => import('./components'));
+const { Layers24 } = lazily(() => import('@konturio/default-icons'));
```
propakov: it doesn't seem to me like a good idea to import @konturio/default-icons lazily
just because we do it in a lot of parts of app not lazily

src/features/layers_panel/index.tsx
```diff
@@ -1,5 +1,14 @@
-import { MapLayerPanel } from './components';
+import { lazily } from 'react-lazily';
+import { i18n } from '~core/localization';
+import { MIN_HEIGHT } from './constants';
+import type { PanelFeatureInterface } from 'types/featuresTypes';
 
-export function MapLayersList() {
-  return <MapLayerPanel />;
-}
+const { PanelContent } = lazily(() => import('./components'));
```
propakov, line 7: it feels like lazy loading (lazily + <Suspense fallback={null}> wrapping) should be done on place of usage (Map.tsx)

src/widgets/LayersAndLegends/components/LayersAndLegends.tsx
```diff
@@ -0,0 +1,90 @@
...
```
propakov:
```suggestion
export function LayersAndLegends({ layersProps = {}, legendProps = {}}: PanelProps) {
  const {
    content: layersPanelContent,
    minHeight: layersMinHeight,
    header: layersHeader,
    panelIcon: layersIcon,
  } = layersProps;
  const {
    content: legendPanelContent,
    minHeight: legendMinHeight,
    header: legendHeader,
    panelIcon: legendIcon,
  } = legendProps;
```

src/features/layers_panel/index.tsx
IlyaIzr: if we export an object from feature (with type of feature interface), it can't be lazy loaded using `lazily` out of the box, we'll need to write a wrapper
but exporting it unconditionally isn't the options
so that's the middle ground

propakov: ah, I see 🤔

src/views/Map/Map.tsx
```diff
@@ -125,8 +124,14 @@ export function MapPage() {
             userModel?.hasFeature(AppFeature.EVENTS_LIST) &&
             userModel?.feeds && <EventListPanel />
           }
-          layers={userModel?.hasFeature(AppFeature.MAP_LAYERS_PANEL) && <MapLayersList />}
-          legend={userModel?.hasFeature(AppFeature.LEGEND_PANEL) && <Legend />}
+          layersAndLegends={
+            <LayersAndLegendsWidget
+              layers={layersPanel}
```
propakov: why not just `userModel?.hasFeature(AppFeature.LEGEND_PANEL) && layersPanel` ? no need for isLegendOn prop then 
same with legend

propakov: in that case you don't need src/widgets/LayersAndLegends/index.tsx

vkozio (approved):

PR #277, opened 2022-12-05
chore(analytics, advanced_analytics): 11670 - decouple disaster links from analytical panel

IlyaIzr:
Links moved from analytics to active event card according to new design
image

### DIFF
--- a/src/features/events_list/components/EventCard/EventCard.module.css
+++ b/src/features/events_list/components/EventCard/EventCard.module.css
@@ -9,7 +9,8 @@
   transition: background-color ease-out 0.1s;
   &.active {
     background-color: var(--accent-weak);
-    box-shadow: 0px 0px 0px 1px var(--accent-strong);
+    box-shadow: 0px 0px 0px 0px var(--accent-strong);
+    border-left: var(--half-unit) solid var(--accent-strong);
   }
   &:focus {
     box-shadow: inset 0px 0px 0px 2px var(--accent-strong);
@@ -28,4 +29,21 @@
   color: var(--text-faint-strong);
   justify-content: space-between;
   width: 100%;
+}
+
+.linkContainer {
+  display: flex;
+  flex-direction: column;
+  gap: var(--unit);
+  font-family: var(--font-family);
+  font-style: normal;
+  font-weight: normal;
+  font-size: 12px;
+  line-height: 16px;
+}
+
+.markdown {
+  & p {
+    margin: 0;
+  }
 }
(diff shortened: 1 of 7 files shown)


src/features/events_list/components/EventCard/EventCard.module.css
```diff
@@ -29,3 +30,20 @@
...
+  flex-direction: column;
+  gap: var(--unit);
+  font-family: var(--font-family);
+  font-style: normal;
+  font-weight: normal;
+  font-size: 12px;
+  line-height: 16px;
+}
+
+.markdown {
```
propakov, line 41: you've added it there, but haven't removed it from AnalyticsData.module.css

src/features/events_list/components/EventCard/EventCard.tsx
```diff
@@ -55,6 +61,22 @@ export function EventCard({
         loss={event.loss}
       />
 
+      {isActive && externalUrls?.length ? (
+        <div>
```
propakov: it seems like this level of div is redundant 
I've removed it and nothing has changed

src/features/events_list/components/EventCard/EventCard.module.css
```diff
@@ -29,3 +30,20 @@
...
 }
+
+.linkContainer {
+  display: flex;
+  flex-direction: column;
+  gap: var(--unit);
+  font-family: var(--font-family);
+  font-style: normal;
+  font-weight: normal;
+  font-size: 12px;
```
propakov: --font-xs

propakov: oh, we also have k-font-caption class in typography that repeats all text styles from here
idea was to have 2 classes .linkContainer {
  display: flex;
  flex-direction: column;
  gap: var(--unit);
}  + k-font-caption 
just to reuse existing styles
but not sure if it's a right idea to use it out of component  🤔 
we have only one such usage in our codebase

propakov (approved):
LGTM
Found some possible moments to improve, pls consider

src/features/events_list/components/EventCard/EventCard.module.css
IlyaIzr: it works but yes it's just string out of context, feels like smth everyone would ask then 'where did it came from?'
image

Akiyamka (approved):

Komzpa (approved):
Thanks for raising the coverage :)

PR #279, opened 2022-12-05
feat(view): 12142 add tos and privacy policy to dn

Akiyamka:
image
image

- Support of always visible route mod added
  - now route have `visibility` property that can be `auto`, 'always' and 'never`
- Support of mdx added
  - Regular link in md fully reload page. To avoid this i need to use <Link /> component - that why mdx needed.  
- New routes and views added
  - /about/privacy
  - /about/cookies (temporary, will be replaced with configuration screen in feature)


src/core/router/routes.tsx
```diff
@@ -31,6 +33,22 @@ export const routerConfig: AppRouterConfig = {
...
+      parentRoute: 'about',
+      visibility: 'always',
+    },
+    {
+      slug: 'cookies',
+      title: 'modes.cookies',
+      icon: <Reports16 />,
+      view: <CookiesPage />,
+      parentRoute: 'about',
+      visibility: 'never',
```
IlyaIzr: `visibility` here relates to visibility in sidebar (you can still see the view of the route by the link), may be smth like `isInNavigation` or `isInSidebar`/`inSidebar` is more accurate

src/features/side_bar/components/SideBar/SideBar.tsx
```diff
@@ -143,7 +157,13 @@ export function SideBar({
...
                     size={route.parentRoute ? 'small-xs' : 'small'}
                     active={currentRoute?.slug === route.slug}
-                    iconBefore={route.icon}
+                    iconBefore={
+                      route.parentRoute ? (
+                        <SmallIconSlot>{route.icon}</SmallIconSlot>
+                      ) : (
+                        route.icon
+                      )
+                    }
```
IlyaIzr, line 129: It's good
Also it can be a component that handles the logic of what icon to return from route
```
<RouteIcon icon={route.icon} parent={route.parentRoute} ...etc/>
```
In case we need to handle cases like empty icon / capital letter icon or smth

src/views/Cookies/Cookies.module.css
```diff
@@ -0,0 +1,53 @@
+.mainWrap {
+  place-content: center;
+  flex: 1;
+  padding-top: 80px;
+  padding-bottom: 80px;
```
IlyaIzr: maybe we need some TODOS for ui library to store values to then implement consistent paddings for all pages

src/views/Cookies/Cookies.module.css
```diff
@@ -0,0 +1,53 @@
...
+}
+
+a,
+.link {
+  color: var(--accent-strong);
+  text-decoration: none;
+}
+
+.content table {
+  margin-top: calc(var(--double-unit) * 4);
```
IlyaIzr: maybe another candidate to extend our offset library values

src/views/Privacy/Privacy.module.css
```diff
@@ -0,0 +1,38 @@
...
+  padding-bottom: 80px;
+  overflow-y: auto;
+}
+
+.content {
+  max-width: 800px;
+  margin-left: auto;
+  margin-right: auto;
+  font-size: 16px;
+  line-height: 24px;
```
IlyaIzr: i think we have `<Text />` for this size

src/features/side_bar/components/SideBar/SideBar.tsx
```diff
@@ -28,17 +29,30 @@ function routeVisibilityChecker(routes: AppRoute[]) {
...
+        const isActive = route.slug === currentRoute.slug;
+        const haveActiveParentRoute = route.parentRoute
+          ? currentRoute?.slug === route.parentRoute
+          : false;
+        const neighbors = route.parentRoute
+          ? Object.keys(routesTree[route.parentRoute])
+          : [];
+        const haveActiveNeighbor = neighbors.includes(currentRoute.slug);
+
+        return isActive || haveActiveParentRoute || haveActiveNeighbor;
```
IlyaIzr: a bit sophisticated (hide nested routes if no selected routes), but that was a business requirement right?

IlyaIzr (approved):
No breaking changes, lgtm, I've marked some minor points, otherwise I'd like @propakov to make sure translations procedure would work with `.md` files (sorry, i've lost the procedure doc somewhere in slack)

src/features/side_bar/components/SideBar/SideBar.tsx
Komzpa: I believe the requirement was that ToS should always be on the screen?

src/views/Cookies/cookies.en.md
```diff
@@ -0,0 +1,45 @@
...
+| \_gid                     | .disaster.ninja   | Registers a unique ID that is used to generate statistical data on how the visitor uses the website.                                                                                                                                       | 24 hour                    | Analytics     |
+| \_yasc                    | .yandex.ru        | Collecting data about the user's interests and displaying relevant advertising on other sites.                                                                                                                                             | 385 days, 10 hours, 21 min | Advertisement |
+| \_ym_d                    | .yandex.ru        | Allows to use additional features of the website and customize it according to users preferences                                                                                                                                           | 349 days, 7 hours, 49 min  | Analytics     |
```
Komzpa: that's metrika? description seems wrong. "Duration" seems arbitrary? @katluk

src/views/Cookies/cookies.en.md
```diff
@@ -0,0 +1,45 @@
+# Cookie files policy and operational data
```
Komzpa: Make sure this one can be translated via normal gettext flow.

Komzpa (changes requested):
Please fix for translations to work using the same mechanism, address suggestions by Ilya (esp. naming). Re-review cookie page content.

Some notes from AI overlords:

In the routerConfig type, the visibility property of the PrivacyPage route is set to always, but the CookiesPage route is set to never. It might make more sense to set the visibility of both routes to the same value, or to use a different property (e.g. showInMenu) to control whether the route appears in the navigation menu.

The title property of the CookiesPage route is currently set to the string "modes.cookies" instead of a translation for the page's title. It would be more user-friendly to use the translation instead, e.g. i18n.t('modes.cookies').

src/core/router/routes.tsx
Akiyamka: The router is a higher-level entity, the sidebar is what displays it. I don't think the sidebar should be mentioned here, it may not be there at all, or there may be more than one. 

I summary i want to say that this option rule the visibility everywhere in app, not only in sidebar (it just current state where we have only sidebar for show this)

src/features/side_bar/components/SideBar/SideBar.tsx
Akiyamka: I think when we have more than one condition and variation we will do so

src/views/Cookies/Cookies.module.css
Akiyamka: Good idea, but I'm not sure if it should be a ui kit or a component in the ninja

src/features/side_bar/components/SideBar/SideBar.tsx
Akiyamka: you are both right, we have both requirements in the design

src/views/Cookies/cookies.en.md
Akiyamka: @Komzpa As far as I know we have no plans to translate the terms, I will clarify this point once again

src/features/side_bar/components/SideBar/SideBar.tsx
```diff
@@ -28,17 +29,30 @@ function routeVisibilityChecker(routes: AppRoute[]) {
...
+      default:
+        // always show top level routes
+        // hide nested routes if no selected routes
+        if (!route.parentRoute) return true;
+        if (currentRoute === null) return false;
+        const isActive = route.slug === currentRoute.slug;
+        const haveActiveParentRoute = route.parentRoute
+          ? currentRoute?.slug === route.parentRoute
+          : false;
+        const neighbors = route.parentRoute
```
propakov: neighbours

src/features/side_bar/components/SmallIconSlot/SmallIconSlot.module.css
```diff
@@ -0,0 +1,8 @@
+.smallIconSlot {
+  width: '24px';
```
propakov:
```suggestion
  width: 24px;
```

src/views/Cookies/cookies.en.md
Akiyamka: > In the routerConfig type, the visibility property of the PrivacyPage route is set to always, but the CookiesPage route is set to never. It might make more sense to set the visibility of both routes to the same value, or to use a different property (e.g. showInMenu) to control whether the route appears in the navigation menu.

a good example that sometimes ai generates nonsense

src/features/side_bar/components/SideBar/SideBar.tsx
```diff
@@ -28,17 +29,30 @@ function routeVisibilityChecker(routes: AppRoute[]) {
...
+          ? Object.keys(routesTree[route.parentRoute])
+          : [];
+        const haveActiveNeighbor = neighbors.includes(currentRoute.slug);
```
propakov, line 52: haveActiveNeighbour

src/views/Cookies/cookies.en.md
Akiyamka: > The title property of the CookiesPage route is currently set to the string "modes.cookies" instead of a translation for the page's title. It would be more user-friendly to use the translation instead, e.g. i18n.t('modes.cookies').

this is done on purpose, since this route is invisible, I do not generate a translation for it

src/views/Privacy/privacyPolicy.en.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+   The Controller provides information on action taken on data subject request without undue delay and in any event within one month of receipt of the request. That period may be extended by two further months where necessary, taking into account the complexity of case and number of the requests. The Controller informs data subject about extension within one month of receipt of the request with the reasons for the delay. The information can be send by email.
+
+## Privacy Policy changes
+
+1. The Controller’s offer may expand in the future. This means The Controller will be obliged or authorized to enter those changes into the Privacy Policy.
+2. New Privacy Policy versions will occur on the website with a proper notification attached.
+3. Every change to The Privacy Policy will be binding from the day it’s being published on the website with proper notification. All changes will be properly announced and marked/featured with a one month notice from the changes implementation within The Privacy Policy.
+
+In case of further questions surrounding The hereby given Privacy Policy, please contact The Controller using his e mail: hello@kontur.io or contact address.
```
propakov: please contact The Controller using his e mail

feels like type his -> this 🤔

propakov:
just a question: 
there is no way on ui to reach about/cookies
is it by requirements? can't find anything in task description about it, sorry  🤷‍♂️

src/views/Cookies/cookies.en.md
Akiyamka: > Please fix for translations to work using the same mechanism

I think document translations require a bit different translation flow. For that we should have <document_name>.<language>.md - then front-end can get necessary file for user language. 

This seems more sensible than trying to fit it into current flow for several reasons.
Let's see what our options are:
- separate document to bunch of small small pieces. Give each piece a key. It will most likely contain a number. This is a lot of manual work, difficult to update, especially if the text structure changes. So it's bad option.
- Put all text into one peace - markdown is sensitive to line breaks, json has them stored only in a serialized form.
serialization - deserialization through multiple formats is a complex and expensive task. As my short research has shown, there are no ready-made tools that are suitable for this
- The last option i see - to store the document for each language separately. The API of our toolkit makes it easy to select the desired one for the current user language and even change it without reloading the page (not done now, because there is no such a need at the moment, but the foundation was laid and thought of how this will be implemented)
The disadvantage of the last option is that the documents will need to be translated separately from the gettext, which is the lesser of evils described above.

What do you think?

src/features/side_bar/components/SideBar/SideBar.tsx
Akiyamka: > Neighbours is an alternative spelling of neighbors. Where neighbors is the preferred spelling in American English, neighbours is the standard in British English.

https://writingexplained.org/neighbours-or-neighbors-difference

Akiyamka:
> just a question:
> there is no way on ui to reach about/cookies
> is it by requirements? can't find anything in task description about it, sorry 

Good catch, it was additional requirement that written me in slack, i will fix it in fibery.
This route available from by direct link from terms doc 
https://github.com/konturio/disaster-ninja-fe/pull/279/files#diff-8611ed9aeb276fa202e16fd0d5590c187a9bdef376ddad075375050cee98612aR61-R63

src/features/side_bar/components/SideBar/SideBar.tsx
propakov: wow. didn't know that!

propakov: beg your pardon, sir 🙏

src/views/Privacy/privacyPolicy.en.mdx
Akiyamka: nice catch, type confirmed and fixed

src/views/Cookies/Cookies.module.css
Akiyamka: I created new component in components/Layout - `Article`

src/views/Cookies/Cookies.module.css
Akiyamka: can't find where such values also used 
image

src/views/Cookies/cookies.en.md
Komzpa: @migrek please review. 

My suggestion is that whatever you invent it ends up in the po/pot files.

src/core/router/routes.tsx
IlyaIzr: yeah, i agree it's not about Sidebar itself, that's why i proposed an option about general navigation. In the end this property defines whether the route is visible in any apps navigation

src/views/Cookies/Cookies.module.css
IlyaIzr: i mean it seems like another arbitrary value

propakov (approved):
👍

src/views/Cookies/cookies.en.md
Akiyamka: Can we resolve it with todo to this task?
https://kontur.fibery.io/search/ty_vT#Tasks/Task/Improve-translation-flow-for-documents-13704

src/views/Cookies/Cookies.module.css
Akiyamka: i just remove that because header already have margin

Komzpa (approved):

PR #280, opened 2022-12-06
fix(cosmos): now you are able to run and build cosmos

propakov:
https://kontur.fibery.io/Tasks/Task/osmos-fix-for-DN-FE-13679


IlyaIzr (approved):
cosmos runs for me locally

PR #282, opened 2022-12-06
fix(bivariate-hexagon-tooltip): 12683 - add missing label processing

propakov:
https://kontur.fibery.io/Tasks/Task/Missing-axes-names-on-the-tooltip-13683
image

### DIFF
--- a/src/components/MapHexTooltip/MapHexTooltip.tsx
+++ b/src/components/MapHexTooltip/MapHexTooltip.tsx
@@ -7,6 +7,7 @@ import {
   LOW,
   MEDIUM,
 } from '~components/BivariateLegend/const';
+import { formatBivariateAxisLabel } from '~utils/bivariate';
 import s from './MapHexTooltip.module.css';
 import type { BivariateLegend } from '~core/logical_layers/types/legends';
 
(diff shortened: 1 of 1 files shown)


src/components/MapHexTooltip/MapHexTooltip.tsx
```diff
@@ -47,8 +48,8 @@ export const MapHexTooltip = ({
 
       <div className={s.labels}>
         <div className={s.column}>
-          <span>{axis.x.label}</span>
-          <span>{axis.y.label}</span>
+          <span>{axis.x.label || formatBivariateAxisLabel(axis.x.quotients)}</span>
+          <span>{axis.y.label || formatBivariateAxisLabel(axis.y.quotients)}</span>
```
Akiyamka, line 52:
```suggestion
          <span>{axis.x.label ?? formatBivariateAxisLabel(axis.x.quotients)}</span>
          <span>{axis.y.label ?? formatBivariateAxisLabel(axis.y.quotients)}</span>
```

Akiyamka (approved):

src/components/MapHexTooltip/MapHexTooltip.tsx
propakov: it comes as an empty string from backend, so it won't work this way 🤷‍♂️

PR #287, opened 2022-12-08
feat(bivariate-renderer): 13219 - hex tooltip moved to map's popup

propakov:
https://kontur.fibery.io/Tasks/Task/The-hex-tooltip-is-sticked-to-the-cursor-while-moving-the-map-by-arrows-on-the-keyboard-13219
Screenshot 2022-12-08 at 14 55 48

### DIFF
--- a/src/core/logical_layers/renderers/BivariateRenderer.tsx
+++ b/src/core/logical_layers/renderers/BivariateRenderer.tsx
@@ -1,7 +1,7 @@
 import throttle from 'lodash/throttle';
 import isNil from 'lodash/isNil';
 import { h3ToGeoBoundary, geoToH3 } from 'h3-js';
-import { Popup } from 'maplibre-gl';
+import { Popup as MapPopup } from 'maplibre-gl';
 import { createRoot } from 'react-dom/client';
 import { LogicalLayerDefaultRenderer } from '~core/logical_layers/renderers/DefaultRenderer';
 import { generateLayerStyleFromBivariateLegend } from '~utils/bivariate/bivariateColorThemeUtils';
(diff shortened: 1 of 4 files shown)


src/core/logical_layers/renderers/BivariateRenderer.tsx
```diff
@@ -1,6 +1,8 @@
 import throttle from 'lodash/throttle';
 import isNil from 'lodash/isNil';
 import { h3ToGeoBoundary, geoToH3 } from 'h3-js';
+import { Popup } from 'maplibre-gl';
```
IlyaIzr: How about
```suggestion
import { Popup as MapPopup } from 'maplibre-gl';
```
or smth
I didn't even knew it has this thing :D

src/core/logical_layers/renderers/BivariateRenderer.tsx
```diff
@@ -224,30 +224,31 @@ export class BivariateRenderer extends LogicalLayerDefaultRenderer {
...
+          axis={legend.axis}
+          values={valuesByQuotients}
+          hexagonColor={rgba}
+        />,
+      );
+
+      if (this._popup) cleanPopup();
+      this._popup = new Popup({
+        closeOnClick: true,
+        className: 'bivariateHexagonPopupContentRoot',
```
IlyaIzr: this className is very specific - so it's weird we add it to global css
can we use it in our usual flow with `import s from 'some.module.css'` and `className={s.bivariateHexPopup}`?

propakov: yeah, i tried with usual flow, but it didn't work
it seems because only top classname is from module.css and all others are not

propakov: aah, I just need to use :global scope

src/core/logical_layers/renderers/BivariateRenderer.tsx
propakov: Nice! I would be more clear that it's not a usual popup. 👌

IlyaIzr:
I'm concerned that this approach allows adding tooltips to the map at least from any layer renderer.
Best theoretical solution is to create single responsible service for it elsewhere in the core. In that case we'll need to think about consequences and design of it (interactions with other tooltip-like stuff)

Least what you should do is to provide cleanup mechanism for introduced functionality - it's missing in the code and in practical usage i can turn off bivariate layer and the popup will still be there.
(Also it feels that closing cross offset needs to be increased)
image

propakov:
@IlyaIzr thanks for this awesome catch! 
I've done with cleanup mechanism, please check it out. 

Regarding your comment:
|   this approach allows adding tooltips to the map at least from any layer renderer.
Only from particular renderer, not every of them. 
I need to handle it on bivariate layers, so do this in bivariate renderer only special way.
Maybe I don't get this point. 🤔 

Agree with you about centralized single responsible service to handle it, but it's whole new entity.
And it's need to be done after we unbound architecture to event-based approach.
Would do it in a separate task for sure as it gonna be pretty big one.

src/core/logical_layers/renderers/BivariateRenderer.tsx
```diff
@@ -332,25 +336,9 @@ export class BivariateRenderer extends LogicalLayerDefaultRenderer {
     ) as MapListener;
 
     map.on('zoom', () => {
-      cleanHover();
-      cleanSelection();
-      currentTooltipAtom.turnOffById.dispatch(TOOLTIP_ID);
+      this.cleanPopupWithDeps(map);
     });
```
IlyaIzr: it's not a subject of this PR, but yeah, another listener that has no cleanup

propakov: yeah, will fix it. nice catch!

propakov: fixed!

IlyaIzr (approved):

PR #285, opened 2022-12-07
fix(sidebar): 13451 - also define app name by app id and change icon in tab

IlyaIzr:
App name and icon (also in tab) defined by app id
also small fix for toggle button styles (that was written off by rebases)
image

Am i putting effect of changing tab icon in the right place? Should component name be affected by that? What do you think?

PR #290, opened 2022-12-09
fix(core: Resizer): 13601 - increase length of analytical panel 

IlyaIzr:
Analytical panel open fully by smart column as was requested
(value of desired height was added to describe it)
image
I've changed logic of `class Resizer ` - previously if panels were to high they were shrinking evenly until they couldn't and then panels were closing in case space was still needed. As a result - we had shrinked and closed panels with free screen space
image
to lessen the damage and improve logic overall now we check if shrinking panels can help at all - if yes - we're shrinking. If no - we're closing and calculating again


Akiyamka:
Short view looks broken, maby not related to this task
image
After switching to short state panel keep size - should we recalculate pannel size after short/long change, or maybe make short view non-resizable?

IlyaIzr:
> Short view looks broken

Yeah, fixed elsewhere

Akiyamka:
image
App logo looks broken in mobile view (on prod it's ok)

IlyaIzr:
> App logo looks broken in mobile view (on prod it's ok)

Happened because backend on dev now gives relative path, and on FE it's properly accepted only here #285

Akiyamka:
image
Look like some miss calculations - in firefox and chrome i have a few pixels scroll that disappear after close-open action

IlyaIzr:
> Look like some miss calculations - in firefox and chrome i have a few pixels scroll that disappear after close-open action

Interesting, i can tweak number for aesthetics! 
In any case the task is to make sure all analytical fields can be seen in open panel without need for resizing

Akiyamka:
> Interesting, i can tweak number for aesthetics! 

Sound like it some hardcode 😅

IlyaIzr:
> Sound like it some hardcode 😅

Yes, it is! Since we have minimum height requirements in pixels i think it's okay

src/core/store/columnContext.tsx
```diff
@@ -14,6 +14,7 @@ export type PanelMeta = {
   closeCb: () => void;
   minHeight: number;
   getOpenState: () => boolean;
+  debug_caller?: string;
```
Akiyamka: What the reason to use camel_case naming here?

Akiyamka: And i can't get how and for what this used

IlyaIzr: to draw attetntion on debug process 
Then i left this key and by doing that i'm asking - should i keep it or remove it?

src/core/store/columnContext.tsx
```diff
@@ -38,15 +39,23 @@ class PanelsRepository {
...
-    return Array.from(this.panels).reduce((acc: [PanelMeta, number][], c) => {
-      const height = c.resizableNode.getBoundingClientRect().height;
-      if (height > c.minHeight) {
-        acc.push([c, height]);
-      }
-      return acc;
-    }, []);
+  getPanelsWithExtraSpace(): [[PanelMeta, number, number][], number] {
+    let totalExtraSpace = 0;
+    /* Returns array with [panel, panelHeight, availableExtraSpace ] entries */
```
Akiyamka: This comment not aligned with `[[PanelMeta, number, number][], number]`

Akiyamka: I think it would be cleaner if you choose one approach:
forEach + variables outside from cycle for accumulate values, or reduce. Currently it's mix of those without visible reason

src/core/store/columnContext.tsx
IlyaIzr: it helps when you debugging cards/panels - they don't have names, only minHeight and Jsx elements

src/core/store/columnContext.tsx
IlyaIzr: Okay, but is it wrong to increment outside variable from reduce callback?

src/core/store/columnContext.tsx
Akiyamka: Don't leave it if that's not used, or implement debug tool fully, or remove it.

src/core/store/columnContext.tsx
Akiyamka: the point of reduce is to give you a function that already has this variable in scope

src/core/store/columnContext.tsx
```diff
@@ -131,16 +140,26 @@ export class Resizer {
     }
   }
 
+  // why not resize first and close only after?
+  // then we have scenario when we shrinked all the panels and still have to close some
+  // after closing free space left but the panels are already shrinked
```
Akiyamka, line 142: Very clear!

src/features/analytics_panel/constants.ts
```diff
@@ -1 +1,2 @@
 export const MIN_HEIGHT = 119;
+export const DESIRED_HEIGHT = 320
```
Akiyamka: DESIRED_HEIGHT need to calculate from content, it dynamic one (326.667px as i can see right now)

src/utils/hooks/useResizer.tsx
```diff
@@ -8,6 +8,7 @@ export const useHeightResizer = (
   setIsOpen: (value: SetStateAction<boolean>) => void,
   isOpen: boolean,
   minHeight: number,
+  debug_caller?: string
```
Akiyamka: https://github.com/konturio/disaster-ninja-fe/pull/290/files#r1045709731

Akiyamka (changes requested):
1) DESIRED_HEIGHT must be calculated from content
2) debug_caller - remove it or add logick that use it. In second case - looks like it's "pannelDebugName" o just "id"
3) use reduce or forEach, instead of use reduce as forEach

src/core/store/columnContext.tsx
IlyaIzr: it's very nice suggestion about just `id`! I wonder how i haven't thought about it (probably because this id doesn't utilized anywhere)

src/core/store/columnContext.tsx
```diff
@@ -162,3 +178,7 @@ export class Resizer {
     });
   }
 }
+
+export function PanelForTest() {
```
Akiyamka: remove it

Akiyamka (approved):

Akiyamka (commented):
Sometimes analytics panel have wrong height

Akiyamka (approved):

PR #294, opened 2022-12-15
fix(bivariate-matrix): 11443 - fix cutted icons of the denominators 

propakov:
https://kontur.fibery.io/Tasks/Task/Cutted-icons-of-the-denominators-in-the-Bivariate-Manager-11443

- removed forwardRef chain and used context for passing callback
- types fixed after forwardRef removed, they were not working
- shifting row/col fixed, matric cell sides are set as particular values
- rotated matrix is wrapped in container with calculated size, and this container is flex-end to the right (that's how matrix jumping fixed, had a discussion there - https://konturio.slack.com/archives/C02DEKP2MKP/p1671023811699569) 
- until calculation finished - matrix is hidden (to prevent first draw jumping)
- small boxes skewed 45/2=22.5 degrees, to help the person to more smoothly follow the bend

Notes: 
src/features/bivariate_manager/components/BivariateMatrixControl/index.tsx diff looks broken in github, please check it locally, there are not so many changes here

**Screenshots:** 
_borders on screenshots are just to show correct matrix positioning
small matrix_
image
big matrix
image
image
image
image

### DIFF
--- a/src/features/bivariate_manager/components/BivariateMatrixContainer/BivariateMatrixContainer.tsx
+++ b/src/features/bivariate_manager/components/BivariateMatrixContainer/BivariateMatrixContainer.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useRef } from 'react';
+import { useEffect, useMemo, useRef } from 'react';
 import clsx from 'clsx';
 import { useAtom } from '@reatom/react';
 import { Text } from '@konturio/ui-kit';
@@ -11,7 +11,10 @@ import { BivariateGreetingsContainer } from '~features/bivariate_manager/compone
 import { i18n } from '~core/localization';
 import { focusedGeometryAtom } from '~core/shared_state';
 import { isGeometryEmpty } from '~core/bivariate';
+import { AXIS_CAPTIONS_WIDTH, MATRIX_SCALE } from '../BivariateMatrixControl/constants';
 import s from './BivariateMatrixContainer.module.css';
+import { BivariateMatrixContext } from './bivariateMatrixContext';
+import type { BivariateMatrixContextInterface } from './bivariateMatrixContext';
 
 interface BivariateMatrixContainerProps {
   className?: string;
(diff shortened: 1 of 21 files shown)


src/features/bivariate_manager/components/BivariateMatrixContainer/BivariateMatrixContainer.tsx
```diff
@@ -57,28 +61,27 @@ const BivariateMatrixContainer = ({ className }: BivariateMatrixContainerProps)
...
-        </div>
+        {statesToComponents({
+          loading: (
+            <div className={s.loadingContainer}>
+              <LoadingSpinner />
+            </div>
+          ),
+          error: () => (
+            <div className={s.errorContainer}>
+              <ErrorMessage message="Unfortunately, we cannot display the matrix. Try refreshing the page or come back later." />
```
Komzpa: is this translatable?

src/features/bivariate_manager/components/BivariateMatrixControl/types.ts
```diff
@@ -1,22 +1,17 @@
...
   quotients: BivariateMatrixQuotientType[];
 };
 
-export type CornerRange =
-  | 'good'
-  | 'bad'
-  | 'important'
-  | 'unimportant'
-  | 'neutral';
+export type CornerRange = 'good' | 'bad' | 'important' | 'unimportant' | 'neutral';
```
Komzpa, line 14: this set is not supposed to be limited by front end?

Komzpa (changes requested):
Screenshots look good. The very initial design however had the small boxes skewed 45/2=22.5 dsegrees, to help the person to more smoothly follow the bend - if that could be done that would be fantastic.

In code, some questions are raised.

Komzpa:
If we can get 7 lines of code covered by tests that will also significantly improve the mood of codecov :)

src/features/bivariate_manager/components/BivariateMatrixContainer/BivariateMatrixContainer.tsx
propakov: Nope, ErrorMessage just do some stylings, will add! thanks for the catch!

src/features/bivariate_manager/components/BivariateMatrixControl/types.ts
propakov: no, this type just describes possible values. 
on frontend we use values as just string, no checks is it a part of set or not.

Akiyamka (approved):

propakov:
@Komzpa test added

propakov:
oh, god, it worked locally :/
will investigate later

Komzpa (approved):

PR #304, opened 2022-12-19
chore(analytics, advanced_analitics, layers_panel, legends_panel, widgets): 12463, 13883 - merge analytics and advanced analytics panels

IlyaIzr:
I'll add image when adv analytics will load 😆 
Changes: 
- I've decided that layersAndLegends widget is almost the same of what I need to merge analytics, so i've created more universal widget
- Suspense wrapper added for lazy imported components -> blinking is fixed
- A simple parameter added to resizing watcher. If we don't want panel to be resized - just set it to false. Accordingly that attribute was added to panel export type
- Min height for advanced analytics was added

P.s. be not afraid of the size: lots of renaming and moving components from one place to another


src/types/featuresTypes.ts
```diff
@@ -3,4 +3,7 @@ export type PanelFeatureInterface = {
   panelIcon?: JSX.Element;
   content?: JSX.Element | null;
   minHeight?: number;
+  maxHeight?: number;
+  // noShrinking?
+  skipAutoResize?: boolean;
```
IlyaIzr, line 8: should it be `noShrinking`?

propakov:
- we have double scroll, difinetely not ok
- container resizer is broken
Screenshot 2022-12-20 at 12 52 14

src/features/analytics_panel/components/PanelContent/PanelContent.module.css
```diff
@@ -0,0 +1,19 @@
...
+}
+
+/* laptop */
+@media screen and (max-width: 1530px) {
+  .panelBody {
+    padding: var(--double-unit);
+    overflow-y: auto;
+  }
+}
+@media screen and (max-width: 960px) {
```
propakov, line 14:
```suggestion
/* mobile */
@media screen and (max-width: 960px) {
```

Komzpa (changes requested):
Double scroll is bad. 
It is ok to have resizer broken, this will get fixed with redesign of Advanced Analytics.

src/widgets/PrimaryAndSecondaryPanel/components/PrimaryAndSecondaryPanelWidget.tsx
```diff
@@ -67,8 +80,10 @@ export function LayersAndLegendsWidget({ layersProps, legendProps }: PanelProps)
         contentClassName={s.contentWrap}
         contentContainerRef={handleRefChange}
         customControls={panelControls}
-        contentHeight={isShort ? 'min-content' : undefined}
+        // TODO look at it
```
propakov: comment feels unclear, need more details what to do

src/features/advanced_analytics_panel/components/PanelContent/PanelContent.module.css
```diff
@@ -0,0 +1,13 @@
+.panelBody {
+  width: 790px;
+  height: 640px;
+  padding: var(--double-unit);
+  resize: both;
+  overflow: auto;
+}
+
+@media screen and (max-width: 960px) {
```
propakov, line 9:
```suggestion
/* mobile */
@media screen and (max-width: 960px) {
```

src/widgets/PrimaryAndSecondaryPanel/components/PrimaryAndSecondaryPanelWidget.tsx
propakov: without details it gives nothing to person not into context

src/widgets/PrimaryAndSecondaryPanel/components/PrimaryAndSecondaryPanelWidget.tsx
```diff
@@ -43,15 +56,15 @@ export function LayersAndLegendsWidget({ layersProps, legendProps }: PanelProps)
 
   useAutoCollapsePanel(isOpen, onPanelClose);
 
-  if (!layersProps && !legendProps) return null;
+  if (!primaryProps && !secondaryProps) return null;
 
-  const header = !layersProps ? legendProps!.header : layersProps.header;
-  const panelIcon = !layersProps ? legendProps!.panelIcon : layersProps.panelIcon;
+  const header = !primaryProps ? secondaryProps!.header : primaryProps.header;
```
propakov: I don't get the idea, why not-null assertion is used here, if type says secondaryProps and its fields are optional.
Why not just secondaryProps?.header in all places?

src/types/featuresTypes.ts
propakov: it passes to `useHeightResizer` with `noResize` name, so we can reuse this one

IlyaIzr:
Sorry, now that i've loaded adv analitics i did fixes:
mobile view:
image
desktop view (top)
image
(scrolled to bottom)
image

propakov:
nice! 
i've just found that resizer still differs from other panels
are we going to fix it? 
Screenshot 2022-12-20 at 15 01 38

other than this resizer looks good!

propakov:
now there is no resizer at all for advanced analytics, at least it doesn't look broken 👍 
Screenshot 2022-12-20 at 15 14 32

propakov (approved):
🚀

Komzpa (approved):

Akiyamka (approved):

PR #317, opened 2022-12-21
feat(i18n): (13770, 13758, 13764, 13790, 13793): i18n

propakov:
new locales 
https://kontur.fibery.io/Tasks/Task/Add-Korean-language-to-Platform-DN-FE-13764
https://kontur.fibery.io/Tasks/Task/Add-Indonesian-language-to-Platform-DN-FE-13758
https://kontur.fibery.io/Tasks/Task/Add-German-language-to-Platform-DN-FE-13770
updates 
https://kontur.fibery.io/Tasks/Task/Update-Arabic-language-on-Platform-DN-FE-13793
https://kontur.fibery.io/Tasks/Task/Update-Spanish-language-on-Platform-DN-FE-13790

_Notes:_ 
1 as translations come with old invalid key **filter_nominator** translated, I've marked it as fuzzy to retranslate correctly next time
```
#: advanced_analytics_data_list##filter_numerator
#, fuzzy
```


src/core/localization/gettext/ar/common.po
```diff
@@ -75,131 +73,6 @@ msgstr "عرض العنوان التفسيري"
...
-#: bivariate##matrix##caption##tooltip##p1
-msgid ""
-"Charts involving the Base and Annex axes help establish a relationship "
-"between two data sets."
-msgstr ""
-"تساعد المخططات التي تتضمن محوري القاعدة والملحق في إنشاء علاقة بين مجموعتي "
-"بيانات."
-
-#: bivariate##matrix##caption##tooltip##li1
-msgid "Annex axis - parameters that we analyse"
```
Komzpa, line 127: why is this shuffled around the file? it looks like we did not pass the files through msgmerge with our original .pot file

propakov: I found that translator applied some formatting - str length + keys order is changed.
But all keys are here.

propakov: And yeah, I didn't run msgmerge, should we? Just not familiar with this tool and we don't have it in our dev chain.

propakov: we don't this step implemented :/
Screenshot 2022-12-21 at 14 56 41

src/core/localization/TranslationService.ts
```diff
@@ -7,6 +7,12 @@ import en_common from './translations/en/common.json';
 import es_common from './translations/es/common.json';
 // @ts-ignore
 import ar_common from './translations/ar/common.json';
+// @ts-ignore
+import ko_common from './translations/ko/common.json';
+// @ts-ignore
+import id_common from './translations/id/common.json';
```
IlyaIzr, line 13: is 'id' a universal code for Indonesian?
i thought it was just an id first, then i thought it's Indian (then i remembered it's called hindi), and then i look at the PR description and found the right answer 😆

propakov: yep, you're right =)

src/core/localization/gettext/ar/common.po
propakov: ah, it was just incorrectly placed first time. 
now first bivaraite key #: bivariate##panel##header goes right after #: drawings##self_directions_not_supported
like in our POT, so it's all correct

Komzpa (approved):

PR #318, opened 2022-12-21
chore(events_list): 10166 - add description for active event and short state event

IlyaIzr:
event description is now provided
image
so we show it for active event in full panel state
image
image

and for short state too
image


src/features/events_list/components/CurrentEvent/CurrentEvent.tsx
```diff
@@ -7,7 +7,13 @@ import { createStateMap } from '~utils/atoms';
 import { EpisodeTimelineToggle } from '../EpisodeTimelineToggle/EpisodeTimelineToggle';
 import { EventCard } from '../EventCard/EventCard';
 
-export function CurrentEvent({ hasTimeline }: { hasTimeline?: boolean }) {
+export function CurrentEvent({
+  hasTimeline,
+  showDescription,
```
propakov:
```suggestion
  hasTimeline = false,
  showDescription = false,
```

propakov (approved):

PR #326, opened 2022-12-22
reset user token after failed checkAuth() method in AuthClient

IlyaIzr:
we have problem when failed call for appId crashes app
that case happened when changed auth service made stored cookie corrupted -> call for appId returned 401 and app crashed
i'm fixing that specific case here


.husky/pre-push
```diff
@@ -11,4 +11,4 @@ if [ $EXIT_CODE == 1 ]; then
 fi
 
 npm run typecheck
-npm run depcruise
+npm run depcruise
```
Akiyamka: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

Akiyamka (approved):

PR #331, opened 2022-12-25
fix(core): 13914 - reset login state if getting 401 from the server fix # 2

IlyaIzr:
what's the problem:
after getting 401 for bad token user bad token is erased but user state doesn't change

how i've fixed that:
idea was to clear user states in `ApiClient.init` `unauthorizedCallback` method
this process was described in `AuthClient` `logout` method
``` ts
public logout() {
    this._apiClient.logout();
    currentUserAtom.setUser.dispatch();
    userStateAtom.logout.dispatch();

    if (window['Intercom']) {
      appConfig.intercom.name = window.konturAppConfig.INTERCOM_DEFAULT_NAME;
      appConfig.intercom['email'] = null;
      window['Intercom']('update', {
        name: appConfig.intercom.name,
        email: appConfig.intercom['email'],
      });
    }
  }
```

however putting this block into `ApiClient.init` `unauthorizedCallback` caused circular dependencies
I've found out that `ApiClient` `expiredTokenCallback` can have needed functionality after `AuthClient.checkAuth()` method fired. So i did small refactor inside `checkAuth` method and made `expiredTokenCallback` public to be able to run it inside `unauthorizedCallback`

### DIFF
--- a/src/core/api_client/apiClient.ts
+++ b/src/core/api_client/apiClient.ts
@@ -42,7 +42,7 @@ export class ApiClient {
   private refreshToken = '';
   private tokenWillExpire: Date | undefined;
   private checkTokenPromise: Promise<boolean> | undefined;
-  private expiredTokenCallback?: () => void;
+  public expiredTokenCallback?: () => void;
 
   /**
    * The Singleton's constructor should always be private to prevent direct
@@ -138,9 +138,9 @@ export class ApiClient {
     return 'Wrong data received!';
   }
 
-  async checkAuth(
+  getLocalAuthToken(
     callback: () => void,
-  ): Promise<{ token: string; refreshToken: string; jwtData: JWTData } | undefined> {
+  ): { token: string; refreshToken: string; jwtData: JWTData } | undefined {
     this.expiredTokenCallback = callback;
     const authStr = localStorage.getItem(LOCALSTORAGE_AUTH_KEY);
     if (authStr) {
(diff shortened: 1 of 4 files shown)


src/core/api_client/apiClient.ts
```diff
@@ -42,7 +42,7 @@ export class ApiClient {
   private refreshToken = '';
   private tokenWillExpire: Date | undefined;
   private checkTokenPromise: Promise<boolean> | undefined;
-  private expiredTokenCallback?: () => void;
+  public expiredTokenCallback?: () => void;
```
propakov, line 45: I would refactor it a bit: 

1 token expiration is responsibility of apiClient, so authClient should know nothing about expiredTokenCallback
so in AuthClient.ts line 111
`const response = this._apiClient.getLocalAuthToken(this.logout);`

and remove onTokenExpired

2 in apiClient.ts 

public expiredTokenCallback -> public authClientLogoutCallback 
because it's not only about token  expiration

add
```
  private onTokenExpired() {
    if (this.authClientLogoutCallback) {
      console.warn('User session has been expired. Logging out.');
      this.authClientLogoutCallback();
    }
  }
```
  
  and would call it everywhere instead of 
  
```
  if (this.expiredTokenCallback) {
    this.expiredTokenCallback();
  }
```
  
  so in the end in apiClientInstance.ts
  
  apiClient.expiredTokenCallback?.(); -> apiClient.authClientLogoutCallback?.();

vkozio (approved):

PR #349, opened 2022-12-30
fix(core): 12770 - count category when sorting layers

IlyaIzr:
Visually layers are also sorted considering their category - 'base', 'overlay' or no category
image
I've updated test and added new to cover new sorting
I've created a set of settings to connect mapbox layers id with their 'parent' backend layers id to retrieve layers categories stored in backend layers settings
@Akiyamka proposed to call backend layers as UI layers and we might use that naming instead. Feel free to argue (for example focused geometry isn't really a backend layer, it's just the easier way for me to grasp logic)

# bug fix example
dn_no_bug_example


src/core/logical_layers/renderers/BivariateRenderer.tsx
```diff
@@ -180,6 +181,7 @@ export class BivariateRenderer extends LogicalLayerDefaultRenderer {
         return;
       }
       const layer = { ...layerStyle, id: layerId, source: this._sourceId };
+      generatedMapboxLayersParents.set.dispatch(layerId, this.id);
       layerByOrder(map, this._layersOrderManager).addAboveLayerWithSameType(
```
propakov, line 182: if it every time goes 2 calls `generatedMapboxLayersParents.set.dispatch `then `layerByOrder. addAboveLayerWithSameType `why not to put `generatedMapboxLayersParents.set` inside `addAboveLayerWithSameType` ?  🤔

src/core/logical_layers/utils/layersOrder/layersOrder.ts
```diff
@@ -43,76 +63,121 @@ export class LayersOrderManager {
...
+    }
+    const givenCategory = givenLayerSettings.category;
+
+    const firstLayerOfNextCategory = mountedLayersOfGivenType?.find((searchingLayer) => {
+      if (givenCategory === 'base') return true;
+      if (givenCategory === 'overlay')
+        return (
+          searchingLayer.category === 'overlay' || searchingLayer.category === undefined
+        );
+      // TODO check
```
propakov: check what?

propakov (changes requested):
It looks like it has a bug.
Steps: 
1 select layers: Esri World Imagery and OSM Inspector: Tagging
you see tagging on top as expected
2 select different photo layer - eox.at 2018 cloudless
3 OSM Inspector: Tagging is no more visible, it's under photo layer
expected: OSM Inspector: Tagging still visible
Jan-04-2023 13-05-34

IlyaIzr:
I couldn't reproduce described bug and requested rechecking from @propakov

propakov:
> I couldn't reproduce described bug and requested rechecking from @propakov

Yep, it was reported from wrong environment, rechecked again, works like a charm.
Sorry for this confusion.

src/core/logical_layers/renderers/BivariateRenderer.tsx
IlyaIzr: initially i didn't do it to specifically show that we create a bind with UI-layers and mapbox layers

i will think if i can optimize it, thanks!

propakov (approved):
👍

Komzpa (approved):

src/core/logical_layers/utils/layersOrder/layersOrder.ts
```diff
@@ -43,76 +63,120 @@ export class LayersOrderManager {
       });
     });
   }
+  _initForTests(
+    layersSettings: Map<string, AsyncState<LayerSettings, Error>>,
+    layersParentsIds: Map<string, string>,
+  ) {
+    this._layersSettings = layersSettings;
+    this._layersParentsIds = layersParentsIds;
+  }
```
Akiyamka: [Don't cheat](https://news.drom.ru/volkswagen-36397.html?page=2).

Correct way to do it:

````ts
 init(map: MapLibre, layersSettingsAtom = layersSettingsAtom, generatedMapboxLayersParents = generatedMapboxLayersParents) {
   this.layersSettingsAtom = layersSettingsAtom;
   this.generatedMapboxLayersParents = generatedMapboxLayersParents;
 }

File: src/core/logical_layers/atoms/generatedMapboxLayers.ts [hunk 4]
```diff
@@ -0,0 +1,9 @@
+import { createMapAtom } from '~utils/atoms/createPrimitives';
+
+type MaplibreLayerId = string;
+type UiLayerId = string;
+
+export const generatedMapboxLayersParents = createMapAtom(
```
Akiyamka: filename: generatedMapboxLayers.ts
atomName: generatedMapboxLayersParents
convention about atom names: must end with `atom` word.
type: Map<MaplibreLayerId, UiLayerId>

So in few lines of code we have
- generated layers
- ui layers
- MapLibre layers
- parents layers 

Are there really 4 different types of layers here?

src/core/logical_layers/utils/layersOrder/layersOrder.ts
```diff
@@ -43,76 +63,120 @@ export class LayersOrderManager {
...
-      .find((t) => bottomLayersOfMountedType.has(t));
-
-    // if there's no higher type mounted - return undefined and mount layer on top
-    if (!firstHigherMountedType) return;
-    // if higher type is mounted, get it's bottom layer's id to mount underneath it
-    return bottomLayersOfMountedType.get(firstHigherMountedType)?.id;
+    const orderedLayers = this._getMountedOrderedLayers(map);
+    if (!orderedLayers?.size) return;
+    const mountedLayersOfGivenType = orderedLayers.get(type);
+    const givenLayerParentKey = this._layersParentsIds?.get(id);
```
Akiyamka, line 139: And why store this id card in the atom, it's not like we're in any way reactive to it

src/core/logical_layers/utils/layersOrder/layersOrder.ts
```diff
@@ -126,37 +190,40 @@ export class LayersOrderManager {
...
-      // set the lowest layer for it's type
-      if (!mountedBottomLayersByType.has(layer.type)) {
-        mountedBottomLayersByType.set(layer.type, layer);
-      }
-      n++;
-    }
-    return mountedBottomLayersByType;
+
+    // this loop will come from bottom to top layers (from backgrounds and rasters to lines and symbols)
+    customLayers.forEach((layer) => {
```
Akiyamka, line 200: There was a reason why I used `while` here 
https://www.measurethat.net/Benchmarks/ShowResult/372430
https://www.measurethat.net/Benchmarks/Show/9739/0/array-loop-vs-foreach-vs-map-vs-while#latest_results_block

Akiyamka: This cycle works with pretty big array and performance here very important

Akiyamka: chromium, on the other hand, is the opposite
https://www.measurethat.net/Benchmarks/Show/9739/0/array-loop-vs-foreach-vs-map-vs-while#latest_results_block

Since the slowest version of firefox is still many times faster than the fastest in chromium, and it is forEach, it's ok to use forEach

src/core/logical_layers/utils/layersOrder/layersOrder.ts
```diff
@@ -43,76 +63,120 @@ export class LayersOrderManager {
...
+    }
+    const givenCategory = givenLayerSettings.category;
+
+    const firstLayerOfNextCategory = mountedLayersOfGivenType?.find((searchingLayer) => {
+      if (givenCategory === 'base') return true;
+      if (givenCategory === 'overlay')
+        return (
+          searchingLayer.category === 'overlay' || searchingLayer.category === undefined
+        );
+      if (givenCategory === undefined) return searchingLayer.category === undefined;
```
Akiyamka: It will be much more better and clear if you select find function per category instead of check if's on every iteration

Akiyamka (changes requested):
- don't use atoms if you not need reactivity
- need more clear naming for variables (look like same things have different names)
- don't create separate methods just for tests, the code should not change its behaviour when tested

Akiyamka (approved):

PR #353, opened 2023-01-03
feat(ui): decrease episides timeline padding

dqunbp:
Change episodes timeline event bottom padding
Bump UI version

### DIFF
--- a/src/features/event_episodes/components/EpisodesTimeline/EpisodesTimeline.tsx
+++ b/src/features/event_episodes/components/EpisodesTimeline/EpisodesTimeline.tsx
@@ -8,6 +8,8 @@ import type { Episode } from '~core/types';
 
 type WithRequired<T, K extends keyof T> = T & { [P in K]-?: T[P] };
 
+const timelineMargins = { axis: 2 };
+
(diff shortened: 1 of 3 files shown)


src/features/event_episodes/components/EpisodesTimeline/EpisodesTimeline.tsx
```diff
@@ -65,6 +65,7 @@ export function EpisodesTimeline({ episodes }: { episodes: Episode[] }) {
         stack={timelineState.settings.stack}
         cluster={timelineState.settings.cluster}
         onSelect={onSelect}
+        margin={{ axis: 2 }}
```
propakov: As margin prop is static I would suggest to move it to constant object out of component, to have static link on it.
Here it would be created new object on every render.

dqunbp: Agree. Actually, we don't have any rendering optimizations for `EpisodesTimeline` component, so I don't worry about renders here

dqunbp: @propakov Fixed

propakov (approved):

PR #357, opened 2023-01-04
feat(bivariate-matrix): 13230, 13231 - indicators selection reimplemented

propakov:
https://kontur.fibery.io/Tasks/Task/The-selected-indicators-are-not-highlighted-in-the-BM-for-the-whole-world-13230
https://kontur.fibery.io/Tasks/Task/The-legend-panel-is-not-updating-after-choosing-the-layer-on-BM-for-the-whole-world-13231
no screenshots as it's just a fix of existing behaviour


Komzpa:
> Merging #357 (b50a8f6) into main (b59b3d0) will decrease coverage by 0.30%.

Can we add a bit of tests to make this change positive? Thanks.

src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
```diff
@@ -69,7 +69,10 @@ export const areaLayersDetailsParamsAtom = createAtom(
         const eventFeed = getUnlistedState(currentEventFeedAtom);
         if (eventFeed) newState.eventFeed = eventFeed.id;
       } else {
-        throw Error('Current geometry not from event, event related layer was selected');
+        // TODO: investigate why we needed to throw Error here before
```
IlyaIzr: because it's not expected to retrieve layer-related layer (for example Event Shape) when event id is not presented

src/features/bivariate_manager/components/ConnectedBivariateMatrix/ConnectedBivariateMatrix.tsx
```diff
@@ -125,7 +127,7 @@ const ConnectedBivariateMatrix = () => {
         debouncedShowTooltip({ x: e.clientX, y: e.clientY });
       }
     },
-    [xGroups, yGroups, setMatrixSelection, debouncedShowTooltip, selectedCell],
```
IlyaIzr, line 128: Did you mean it?
i'm asking since `selectedCellRef` is still in the callback

propakov: yep, as selectedCellRef is ref, that is not needed to be synced

src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
propakov: this one breaks overlay selection after you remove selected geometry
Screenshot 2023-01-09 at 12 47 09

IlyaIzr: that's the bug it was hiding then 😄

propakov: so what to do with this line in PR?
I can't find any place when this error catched, so maybe it's ok to stay it as console.error?

Akiyamka: @IlyaIzr you can throw errors only if you somewere catch them. Never throw just for

IlyaIzr (approved):
described bugs fixed

PR #363, opened 2023-01-09
fix(core): 14176 - broken url routing for all pages except the default

IlyaIzr:
We had an issue where redirect started to work before we got list of features
So that blocked user to share routes that require features

### DIFF
--- a/src/core/app/features.ts
+++ b/src/core/app/features.ts
@@ -1,5 +1,7 @@
 import { appConfig } from '~core/app_config';
 import { featureFlagsAtom } from '~core/shared_state';
+import { createBooleanAtom } from '~utils/atoms';
+import { store } from '~core/store/store';
 import type { BackendFeature } from '~core/auth/types';
 import type { ApiClient } from '~core/api_client';
 
@@ -15,6 +17,8 @@ export function loadFeatures(
   return featuresResponse;
 }
 
+export const featuresWereSetAtom = createBooleanAtom(false);
+
 export function setFeatures(value: BackendFeature[] | null) {
   // FIXME: investigate proper app and user feature merge
   const newFeatures = {};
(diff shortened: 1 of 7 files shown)


src/views/CommonView.tsx
```diff
@@ -27,7 +27,8 @@ export function CommonView({
   getAbsoluteRoute: (path: string) => string;
 }>) {
   const [featureFlags] = useAtom(featureFlagsAtom);
-  const [currentLanguage] = useAtom(currentLanguageAtom);
+  // do we need to use this atom here?
```
Akiyamka: We need to activate the languages somewhere, any suggestions for a better location?

src/core/app/features.ts
```diff
@@ -22,4 +25,5 @@ export function setFeatures(value: BackendFeature[] | null) {
     newFeatures[ft.name] = true;
   });
   featureFlagsAtom.set.dispatch(newFeatures);
+  featuresWereAppliedAtom.setTrue.dispatch();
```
Akiyamka: 1. When you need dispatch few actions - batch it.
```
     store.dispatch([
          featureFlagsAtom.set(newFeatures),
          featuresWereAppliedAtom.setTrue(),
        ]);
   ```
   
2. the function now does not simply set features, function name must be changed according to this changes

vkozio:
This problem can be fixed by lazy loading Router in https://github.com/konturio/disaster-ninja-fe/blob/main/src/App.tsx#L7-L8
// import { Router } from '~core/router';
const { Router } = lazily(() => import('~core/router'));

We will rework routing, migrate to different lib, and this simple fix is way to go at the moment.
@Akiyamka thoughts?

Akiyamka:
On the one hand, this could (potentially) slow down the start of the application, but on the other hand, it ensures that the application is initialized sooner. Regardless of the trade-off, I think this is a good solution because it is reliable and transparent

IlyaIzr:
reopen to check if it works on stage as #368 not working on stage build

src/views/CommonView.tsx
IlyaIzr: temp resolving

IlyaIzr: not yet, just removing unused `currentLanguage` variable

src/core/app/features.ts
IlyaIzr: It sets features and stores that it did. I'm not sure i need to change function name.
I've changed atom name that stores it

PR #366, opened 2023-01-10
12087 move tooltip to UI kit part 1

Akiyamka:
This is the first step to bring the tooltip into the kit.
I have separated all platform logic in tooltip/index, the rest of the files do not have any dependencies on parts of the disaster ninja. Functionality the same, nothing changed. Should be more readable (for my taste at least : ) ).  The next step is to put the tooltip in the ui kit package will be in another MR


src/core/shared_state/currentTooltip.ts
```diff
@@ -1,7 +1,12 @@
 import { createAtom } from '~utils/atoms';
 
-export type Position = 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right';
-export type Coords = { x: number; y: number; predefinedPosition?: Position };
+export type TooltipCorner = 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right';
+export type Coords = {
+  x: number;
+  y: number;
+  // TODO: rename to `corner`
+  predefinedPosition?: TooltipCorner;
```
IlyaIzr: we already have prototypes where tooltips have something like 'center-top' position so i'm afraid corners if not enough

dqunbp: Agree with @IlyaIzr, but adding more positions lays out this task scope.

My namings suggestion 
```ts
export type Position = {
  x: number;
  y: number;
  placement?: TooltipPlacement;
};
```
In the future, we will be able to add more placements if needed

dqunbp: Another suggestion - decouple `Coords` and `Position` to two separate `TooltipData` options

src/features/tooltip/TooltipAnchor/TooltipAnchor.tsx
```diff
@@ -0,0 +1,29 @@
...
+  children,
+  getCorner = findTooltipCorner,
+}: PropsWithChildren<{
+  position: TooltipCoords;
+  onClick: (e: React.MouseEvent<HTMLDivElement, MouseEvent>) => void;
+  getCorner?: TooltipCorner | CornerFn;
+}>) {
+  const corner = useCorner(getCorner, position);
+  const top = position.y;
+  const right = (globalThis.visualViewport?.width ?? 0) - position.x;
```
dqunbp: Should the `rigth` be 0 if `globalThis.visualViewport?.width` is nullable?

dqunbp (approved):

PR #372, opened 2023-01-12
fix(bivariate-matrix): 14215 - bivariateStatisticsResourceAtom loadin…

propakov:
…g without geometry fix

https://kontur.fibery.io/Tasks/Task/Matrix-not-loading-after-deleting-geometry-and-page-refresh-14215

Resource atoms don't run if dependency is null. 
In atom's language it means that resource is waiting for input from dependency to run. 
That's why in a lot of AsyncAtoms dependencies accumulated in separate atom. 
Check currentEventResourceAtom and editableLayersListResource.
So in this PR we just provide dependency as not null object, so resource atom runs correctly.

### DIFF
--- a/src/features/bivariate_manager/atoms/bivariateStatisticsResource.ts
+++ b/src/features/bivariate_manager/atoms/bivariateStatisticsResource.ts
@@ -1,3 +1,4 @@
+import { createAtom } from '@reatom/core';
 import { createAsyncAtom } from '~utils/atoms/createAsyncAtom';
 import { apiClient } from '~core/apiClientInstance';
 import { focusedGeometryAtom } from '~core/shared_state';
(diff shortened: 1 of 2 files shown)


IlyaIzr:
# So it works
I miss the description why exactly you've changed dependency for resource atom
As dependency atom of focused geometry had state `FocusedGeometry | null` and now it's `{ focusedGeometry: FocusedGeometry | null; }

Akiyamka (approved):

propakov:
> # So it works
> I miss the description why exactly you've changed dependency for resource atom As dependency atom of focused geometry had state `FocusedGeometry | null` and now it's `{ focusedGeometry: FocusedGeometry | null; }

sorry, I've updated PR description with some context!  @IlyaIzr

PR #374, opened 2023-01-12
feat(analytics-panel): 13921 - Full and short states swap on Analytic…

propakov:
…s panel, made short state as initial

https://kontur.fibery.io/Tasks/Task/Full-and-short-states-are-swapped-on-Analytics-panel-13921

1 swapped short and full states on analytics panel
2 short state is default if feature is on

before
Screenshot 2023-01-12 at 17 06 30

after
Screenshot 2023-01-12 at 17 05 51


Akiyamka:
Tested in cases:
- when only ADVANCED_ANALYTICS_PANEL feature flag enbaled
- when only ANALYTICS_PANEL  feature flag enbaled
- when both enabled

Work fine!

The logic seems a bit distributed, and have extra entities (primary/secondary props, instead of full/short).  It would be better to concentrate it in one place, e.g.

```tsx
const { shortState, fullState } = createAnalyticsPanelStates(featureFlags);

<FullAndShortStatesPanelWidget
  shortState={shortState}
  fullState={fullState}
/>
```

But it out of scope of the task

Akiyamka (approved):

propakov:
@Akiyamka thank you so much for the review! Your suggested refactoring is applied.

PR #376, opened 2023-01-12
fix(events_list): 13936 server error shown after multiple clicks on event card

Akiyamka:
Additionally I fix case with errors when user switch event very fast

image

### DIFF
--- a/src/core/api_client/apiClient.ts
+++ b/src/core/api_client/apiClient.ts
@@ -253,7 +253,7 @@ export class ApiClient {
 
     if (problem.kind === 'canceled') {
       console.debug('Request was canceled');
-    } else if (errorsConfig === undefined || errorsConfig.dontShowErrors !== true) {
+    } else if (errorsConfig === undefined || errorsConfig.hideErrors !== true) {
       this.notificationService.error({
         title: this.translationService.t('error'),
         description: this.translationService.t(errorMessage),
--- a/src/core/api_client/apiClient.ts
+++ b/src/core/api_client/apiClient.ts
@@ -251,14 +251,15 @@ export class ApiClient {
       errorMessage = ApiClient.parseError(problem);
     }
 
-    if (!errorsConfig || !errorsConfig.dontShowErrors) {
-      if (problem.kind !== 'canceled') {
+    if (problem.kind !== 'canceled') {
+      if (!errorsConfig || !errorsConfig.dontShowErrors) {
         this.notificationService.error({
           title: this.translationService.t('error'),
           description: this.translationService.t(errorMessage),
         });
       }
-      console.debug(`Request was canceled`);
+    } else {
+      console.debug('Request was canceled');
     }
 
     throw new ApiClientError(errorMessage, problem);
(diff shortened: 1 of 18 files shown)


src/core/api_client/apiClient.ts
```diff
@@ -252,11 +251,15 @@ export class ApiClient {
       errorMessage = ApiClient.parseError(problem);
     }
 
-    if (!errorsConfig || !errorsConfig.dontShowErrors) {
-      this.notificationService.error({
-        title: this.translationService.t('error'),
-        description: this.translationService.t(errorMessage),
-      });
+    if (problem.kind !== 'canceled') {
+      if (!errorsConfig || !errorsConfig.dontShowErrors) {
```
dqunbp: My suggestion to make this check more readble
```ts
if (!errorsConfig?.dontShowErrors) {
```

dqunbp: One more suggestion)
```ts
if (problem.kind === 'canceled') {
  console.debug('Request was canceled');
} else if (!errorsConfig?.dontShowErrors) {
  this.notificationService.error({
    title: this.translationService.t('error'),
    description: this.translationService.t(errorMessage),
  });
}
```

src/features/events_list/components/FullState/FullState.tsx
```diff
@@ -53,6 +53,17 @@ export function FullState({
     }
   }, [currentEventId, eventsList, virtuoso, setHasUnlistedEvent]);
 
+  const currentEventRef = useRef(currentEventId);
+  currentEventRef.current = currentEventId;
```
dqunbp, line 57: Why do we need this assignment?

src/features/events_list/components/FullState/FullState.tsx
```diff
@@ -53,6 +53,17 @@ export function FullState({
...
+  const eventClickHandler = useCallback(
+    (id: string) => {
+      if (id !== currentEventRef.current) {
+        onCurrentChange(id);
+      }
+    },
+    [onCurrentChange, currentEventRef],
```
dqunbp, line 64: Do we need to include it in deps array? 

`currentEventRef` is immutable

src/features/events_list/components/FullState/FullState.tsx
dqunbp: For assing on each render, not initial only. Got it

src/features/events_list/components/FullState/FullState.tsx
Akiyamka: actually is no difference are we include refs in deps array or not, but linter more happy when we add =)

src/core/api_client/apiClient.ts
Akiyamka: Love it, i make it event more readable after thinking a while
```ts
    if (problem.kind === 'canceled') {
      console.debug('Request was canceled');
    } else if (errorsConfig === undefined || errorsConfig.hideErrors !== true) {
      this.notificationService.error({
        title: this.translationService.t('error'),
        description: this.translationService.t(errorMessage),
      });
    }
```

dqunbp (approved):

PR #379, opened 2023-01-13
chore(core): 13682 - change layouts and styles for footer

IlyaIzr:
Changes in design according to https://kontur.fibery.io/Tasks/Task/Reorder-modes-on-sidebar-14007
I've added extra row in layouts grids to implement design requirements and not let columns overflow page bottom

desktop:
image

laptop:
image

mobile:
image

### DIFF
--- a/src/views/Map/Layouts/Mobile/Mobile.tsx
+++ b/src/views/Map/Layouts/Mobile/Mobile.tsx
@@ -9,7 +9,7 @@ export function MobileLayout({ firstColumn, mapColumn, footer, drawToolbox }) {
         <div className={s.mapSpaceRight}>{mapColumn}</div>
       </div>
       <div className={s.footerWrap}>{footer}</div>
-      <div className={s.fakeFooter}></div>
+      <div className={s.bottomBorder}></div>
     </div>
   );
 }
--- a/src/views/Map/Layouts/Mobile/Mobile.tsx
+++ b/src/views/Map/Layouts/Mobile/Mobile.tsx
@@ -9,6 +9,7 @@ export function MobileLayout({ firstColumn, mapColumn, footer, drawToolbox }) {
         <div className={s.mapSpaceRight}>{mapColumn}</div>
       </div>
       <div className={s.footerWrap}>{footer}</div>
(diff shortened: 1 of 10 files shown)


alesiahil:
1. Desktop. Can you reduce the ruler width so that it fits into the right column?
2. All screens. Can you reduce the ruler height so that it has the same height as Kontur logo does?
3. All screens. Please, make a horizontally centered alignment for the ruler, OSM copyright, Kontur logo and Intercom. It would be better than it is on my layouts. And make sure that 'i' icon on mobile is centered with other elements as well. Because Intercom is 38 px on my layouts but it actually is 28 px on prod and it would be better to center it with other elements as it's smaller on prod. And the distance between ruler and tools or timeline should be at least 8 px.
image

IlyaIzr:
changes

desktop:
image


laptop:
image


mobile:
image
@alesiahil

alesiahil:
It's not critical but:
4. Mobile. Elements are not centered. You need to move ruler up on 6px, i icon - on 8px, kontur logo - on 7px.
5. Desktop. Ruler still does not fit into the right column. Make its width less or if there is no possibility move the bottom side of tools/timeline up so that the bottom line is the same as on the right column.
image

IlyaIzr:
style fixes:
Desktop - moved map tooks higher and restored scale bar width a bit:
image
Mobile - vertically aligned map controls and logo with intercom as requested:
image

alesiahil (approved):

dqunbp (approved):

src/views/Map/Layouts/Mobile/Mobile.tsx
```diff
@@ -9,6 +9,7 @@ export function MobileLayout({ firstColumn, mapColumn, footer, drawToolbox }) {
         <div className={s.mapSpaceRight}>{mapColumn}</div>
       </div>
       <div className={s.footerWrap}>{footer}</div>
+      <div className={s.fakeFooter}></div>
```
Akiyamka: How about giving it a more meaningful name?
A "fake footer" doesn't say why this needed, and what doing, only that it's fake.

PR #386, opened 2023-01-16
fix(logical_layers): 14262 download geojson instead of geojson source

Akiyamka:
Small fix of bug discovered in this tread
https://konturio.slack.com/archives/C01TZENFM63/p1673860512671969?thread_ts=1670222258.104379&cid=C01TZENFM63

This button right now download mapbox geojson source instead of geojson itself 
image

(Fix in 171-180 lines, all other changes it prettier formatting)

### DIFF
--- a/src/core/logical_layers/utils/logicalLayerFabric.ts
+++ b/src/core/logical_layers/utils/logicalLayerFabric.ts
@@ -14,10 +14,7 @@ import { layersMenusAtom } from '../atoms/layersMenus';
 import { deepFreeze } from './deepFreeze';
 import { getMutualExcludedActions } from './getMutualExcludedActions';
 import type { LayerRegistryAtom } from '../types/registry';
-import type {
-  LogicalLayerActions,
-  LogicalLayerState,
-} from '../types/logicalLayer';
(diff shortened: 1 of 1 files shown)


src/core/logical_layers/utils/logicalLayerFabric.ts
```diff
@@ -180,12 +168,16 @@ export function createLogicalLayerAtom(
...
+          if (state.source.source.type === 'geojson') {
+            downloadObject(
+              state.source.source.data,
+              `${
+                state.settings?.name || state.id || 'Disaster Ninja map layer'
+              }-${new Date().toISOString()}.json`,
+            );
+          } else {
+            console.error('Only geojson layers can be downloaded');
+          }
```
IlyaIzr, line 180: I don't know should we tell it to user via toast?

IlyaIzr (approved):

src/core/logical_layers/utils/logicalLayerFabric.ts
Akiyamka: No, for users we hide download buttons, this check mostly for typescript and debug

PR #387, opened 2023-01-16
refactor(common): 13882 migration from deprecated styles to fresh

Akiyamka:
We have update of UI kit design - now we have much more header styles and it separate component.
This commit migrate codebase to new <Header /> component


src/features/user_profile/components/LoginForm/LoginForm.tsx
```diff
@@ -98,7 +98,7 @@ export function LoginForm() {
           <LoadingSpinner message={i18n.t('login.logging_in')} />
         </div>
       )}
-      <Text type="heading-xl">{i18n.t('login.log_in')}</Text>
+      <Heading type="heading-06">{i18n.t('login.log_in')}</Heading>
```
propakov: hm, it differs a bit from what we had. 
is it ok? 
before it was closer to the middle of the screen 
image

propakov: before login inputs were closer to the middle of the screen 
Screenshot 2023-01-16 at 18 02 53

src/features/user_profile/components/SettingsForm/SettingsForm.tsx
```diff
@@ -113,8 +113,9 @@ export function SettingsForm() {
         <div className={s.notFlexParent}>
           <div className={s.flexWrap}>
             <div className={s.profileWrap}>
-              <Text type="heading-xl">{i18n.t('profile.profileSettingsHeader')}</Text>
-
+              <Heading type="heading-06">
```
propakov: same moment here, padding differs drastically
not sure, but is it ok?

Akiyamka: hmm, thanks, i will recheck this

propakov (approved):
apart from heading-06 paddings code LGTM. 
so only this moment need to be checked. 🚀

src/features/user_profile/components/SettingsForm/SettingsForm.tsx
Akiyamka: fixed

src/features/user_profile/components/LoginForm/LoginForm.tsx
Akiyamka: fixed

PR #388, opened 2023-01-16
fix(core): 13914 Reset login state if getting 401 from the server

vkozio:
https://kontur.fibery.io/Tasks/Task/Reset-login-state-if-getting-401-from-the-server-13914
full reload after login and logout
fixes for apiclient
removed some unused code

### DIFF
--- a/src/core/api_client/tests/authorization.test.ts
+++ b/src/core/api_client/tests/authorization.test.ts
@@ -133,6 +133,7 @@ test('refreshToken called when token is expired', async ({ ctx }) => {
 
   // set private field token with new token
   const apiClientObj = ctx.apiClient as any;
+  // token expires in 4 minutes
   apiClientObj.tokenExpirationDate = new Date(new Date().getTime() + 1000 * 60 * 4);
   apiClientObj.token = ctx.token;
 
--- a/src/core/api_client/tests/authorization.test.ts
+++ b/src/core/api_client/tests/authorization.test.ts
@@ -133,7 +133,7 @@ test('refreshToken called when token is expired', async ({ ctx }) => {
 
   // set private field token with new token
   const apiClientObj = ctx.apiClient as any;
-  apiClientObj.tokenWillExpire = new Date(new Date().getTime() + 1000 * 60 * 4);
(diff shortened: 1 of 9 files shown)


src/core/api_client/tests/authorization.test.ts
```diff
@@ -133,7 +133,7 @@ test('refreshToken called when token is expired', async ({ ctx }) => {
 
   // set private field token with new token
   const apiClientObj = ctx.apiClient as any;
-  apiClientObj.tokenWillExpire = new Date(new Date().getTime() + 1000 * 60 * 4);
+  apiClientObj.tokenExpirationDate = new Date(new Date().getTime() + 1000 * 60 * 4);
```
Komzpa, line 137: Can any comment on magic number be added?

src/core/auth/client/AuthClient.ts
```diff
@@ -85,22 +87,24 @@ export class AuthClient {
     const response = await this._apiClient.login(user, password);
     if (response && typeof response === 'object' && 'token' in response) {
       await this.processAuthResponse(response);
+      // reload to init with authenticated config and profile
+      location.reload();
       return true;
```
IlyaIzr, line 92: Do you think expired token can caught user surprised and break it's user flow?

src/core/auth/client/AuthClient.ts
```diff
@@ -85,22 +87,24 @@ export class AuthClient {
...
     }
     return response;
   }
 
   public async checkAuth(): Promise<boolean> {
     try {
-      const response = this._apiClient.getLocalAuthToken(this.onTokenExpired);
+      const response = this._apiClient.getLocalAuthToken(() => this.onTokenExpired());
```
IlyaIzr, line 99: why did you wrap it in arrow function?

vkozio: to bind this

src/core/auth/client/AuthClient.ts
vkozio: yes but it's expected, smoother experience will be implemented in other tasks

IlyaIzr (approved):
questionable stuff was discussed in threads

PR #391, opened 2023-01-17
fix(core): 13454 Fix rounding of coordinates in URL to dynamic precision

vkozio:
https://kontur.fibery.io/Tasks/Task/Fix-rounding-of-coordinates-in-URL-to-dynamic-precision-13454

### DIFF
--- a/src/core/url_store/encoder.ts
+++ b/src/core/url_store/encoder.ts
@@ -1,11 +1,24 @@
+import { URL_ZOOM_OFFSET } from '~core/constants';
 import { URLDataInSearchEncoder } from './dataInURLEncoder';
 
 export const urlEncoder = new URLDataInSearchEncoder({
   order: ['map', 'app', 'event', 'feed', 'layers'], // Doc #1168
   transformers: {
     map: {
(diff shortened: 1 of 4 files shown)


src/core/url_store/encoder.ts
```diff
@@ -1,11 +1,24 @@
...
 
 export const urlEncoder = new URLDataInSearchEncoder({
   order: ['map', 'app', 'event', 'feed', 'layers'], // Doc #1168
   transformers: {
     map: {
-      decode: (str: string) => str.split('/').map((s) => Number(s)),
-      encode: (position: [number, number, number]) => position.join('/'),
+      decode: (str: string) => {
+        const position = str.split('/').map((s) => Number(s));
+        position[0] = position[0] - URL_ZOOM_OFFSET;
```
IlyaIzr: more self-explainatory 
```suggestion
        const zoom = position[0] - URL_ZOOM_OFFSET
        position[0] = zoom;
```

IlyaIzr (approved):
applying same algorithm for precision as in https://github.com/mlevans/leaflet-hash/blob/master/leaflet-hash.js#L41 as was asked in the task description

PR #402, opened 2023-01-20
fix(core):13708 layersInAreaAndEventLayerResource metric is sent only after refresh the page

vkozio:
https://kontur.fibery.io/Tasks/Task/layersInAreaAndEventLayerResource-metric-is-sent-only-after-refresh-the-page-13708

metrics is stopped when page is hidden

additional improvements:
removed currentApplicationAtom and it usages (we have appId in appConfig)
added more missing names for atoms

### DIFF
--- a/src/core/app/init.ts
+++ b/src/core/app/init.ts
@@ -3,7 +3,6 @@ import { appConfig, updateAppConfig } from '~core/app_config';
 import { apiClient } from '~core/apiClientInstance';
 import { urlEncoder, urlStoreAtom } from '~core/url_store';
 import { authClientInstance } from '~core/authClientInstance';
-import { currentApplicationAtom } from '~core/shared_state';
 import { onLogin, onLogout, onPublicLogin } from './authHooks';
 import { runAtom } from './index';
 import type { UrlData } from '~core/url_store';
@@ -40,9 +39,6 @@ export async function appInit() {
   }
 
   updateAppConfig(appInfo);
-
-  // FIXME: refactor remove currentApplicationAtom dependency in other atoms
-  currentApplicationAtom.set.dispatch(initialState.app);
 
   postAppInit(initialState);
 }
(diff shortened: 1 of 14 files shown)


src/core/app/init.ts
```diff
@@ -41,9 +40,6 @@ export async function appInit() {
 
   updateAppConfig(appInfo);
 
-  // FIXME: refactor remove currentApplicationAtom dependency in other atoms
-  currentApplicationAtom.set.dispatch(initialState.app);
```
propakov, line 45: currentApplicationAtom still stays in src/features/create_layer/readme.md
need to remove from docs as it doesn't exist

src/features/create_layer/atoms/editableLayersListResource.ts
```diff
@@ -1,42 +1,15 @@
...
-);
-
 export const editableLayersListResource = createAsyncAtom(
-  editableLayersListDependencyAtom,
-  async (params, abortController) => {
-    if (!params?.appId) return null;
-
+  null,
+  async (_, abortController) => {
     const responseData = await apiClient.post<EditableLayers[]>(
```
propakov, line 10: should we also add it here? `if (!appConfig.id) return null;`

vkozio: appConfig.id cannot be empty

src/core/app/init.ts
vkozio: good point

propakov (approved):

src/features/create_layer/atoms/editableLayersListResource.ts
propakov: so why to handle it here? 
https://github.com/konturio/disaster-ninja-fe/pull/402/files#diff-f926c1c66cd1d38ab061e5691b47a9059b0831317f6a5fe7ab9a7da93e8c9ac4R10

vkozio: fixed

PR #404, opened 2023-01-23
feat(panel): 14327 - Arrow in the "Analytics" tab is in the incorrect…

propakov:
… position at the initiate state

https://kontur.fibery.io/Tasks/Task/Arrow-in-the-Analytics-tab-is-in-the-incorrect-position-at-the-initiate-state-14327

image

### DIFF
--- a/src/utils/hooks/useShortPanelState.tsx
+++ b/src/utils/hooks/useShortPanelState.tsx
@@ -48,9 +48,11 @@ export const useShortPanelState = (props?: UseShortPanelStateProps) => {
   if (skipShortState) {
     const singleControl: PanelCustomControl[] = [
       {
-        icon: panelState === 'full' ? <ChevronUp24 /> : <ChevronDown24 />,
-        onWrapperClick: () =>
-          setPanelState((prevState) => (prevState === 'closed' ? 'full' : 'closed')),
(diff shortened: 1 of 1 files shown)


src/utils/hooks/useShortPanelState.tsx
```diff
@@ -48,9 +48,11 @@ export const useShortPanelState = (props?: UseShortPanelStateProps) => {
   if (skipShortState) {
     const singleControl: PanelCustomControl[] = [
       {
-        icon: panelState === 'full' ? <ChevronUp24 /> : <ChevronDown24 />,
-        onWrapperClick: () =>
-          setPanelState((prevState) => (prevState === 'closed' ? 'full' : 'closed')),
+        icon: panelState === initialState ? <ChevronUp24 /> : <ChevronDown24 />,
```
dqunbp: @propakov 
Can't reproduce on DN prod.
If we have `initialState=closed`,  so we will have closed panel with `ChevronUp24`
Is it correct?

IlyaIzr (approved):

src/utils/hooks/useShortPanelState.tsx
propakov: Nope, it was a mistake. 
Thank you, nice catch!  I've pushed a fix!

PR #407, opened 2023-01-24
fix(layers_in_area): 13152 selected area differs from event shape

Akiyamka:
When use change selected event in order:
1. Select event that was never selected before
2. Select previously active event 

This creates a situation in which all the necessary layers are already in the cache so we don't need to trigger api request (areaLayersDetailsResourceAtom). But we have an old fix in areaLayersLegendsAndSources that listen resource atom changes, and then run imperative update of eventShape layers with new geometry (this workaround was created because eventShape always have same id, but in fact different geometry depending from event in request)
I create task for refactoring that will allow us remove this workaround (14387), and as temporary solution I trigger resource atom with skip 'true` flag - it still not send request to backend, but at the same time trigger listeners of resource atom.

Additional small changes
- ~~improvements in ts config that fixes auto imports of type definitions~~ - blocked by [reatom](https://t.me/reatom_ru/24832) bug
- function name misspelling fixed

### DIFF
--- a/src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
+++ b/src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
@@ -1,11 +1,12 @@
 import { createAtom } from '~utils/atoms/createPrimitives';
 import { enabledLayersAtom } from '~core/logical_layers/atoms/enabledLayers';
-import { focusedGeometryAtom, getEvendId } from '~core/shared_state/focusedGeometry';
+import { focusedGeometryAtom, getEventId } from '~core/shared_state/focusedGeometry';
 import { currentEventFeedAtom } from '~core/shared_state';
 import { layersGlobalResource } from '../layersGlobalResource';
 import { layersInAreaAndEventLayerResource } from '../layersInAreaAndEventLayerResource';
 import { areaLayersDetailsResourceAtomCache } from './areaLayersDetailsResourceAtomCache';
 import type { DetailsRequestParams } from './types';
+import type { LayerInArea, LayerInAreaDetails } from '~features/layers_in_area/types';
 
 export const areaLayersDetailsParamsAtom = createAtom(
   {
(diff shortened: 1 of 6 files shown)


src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
```diff
@@ -1,11 +1,12 @@
 import { createAtom } from '~utils/atoms/createPrimitives';
 import { enabledLayersAtom } from '~core/logical_layers/atoms/enabledLayers';
-import { focusedGeometryAtom, getEvendId } from '~core/shared_state/focusedGeometry';
+import { focusedGeometryAtom, getEventId } from '~core/shared_state/focusedGeometry';
 import { currentEventFeedAtom } from '~core/shared_state';
 import { layersGlobalResource } from '../layersGlobalResource';
 import { layersInAreaAndEventLayerResource } from '../layersInAreaAndEventLayerResource';
 import { areaLayersDetailsResourceAtomCache } from './areaLayersDetailsResourceAtomCache';
 import type { DetailsRequestParams } from './types';
+import type { LayerInArea, LayerInAreaDetails } from '~features/layers_in_area/types';
```
propakov: types are redundant, need to remove

src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsResourceAtom.ts
```diff
@@ -8,6 +8,7 @@ export const areaLayersDetailsResourceAtom = createAsyncAtom(
   areaLayersDetailsParamsAtom,
   async (params, abortController) => {
     if (params === null) return null;
```
propakov: Not your change, but it looks like this line is no more needed as params can't be null, isn't it?
When it's a null before init resource atom doesn't run, and after it can't become null it seems.

src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsResourceAtom.ts
```diff
@@ -8,6 +8,7 @@ export const areaLayersDetailsResourceAtom = createAsyncAtom(
...
+    if (params.skip) return []; // When all 100% layers in cache we still need to trigger ResourceAtomState to run listeners
```
propakov, line 10: after it returns [], in onSuccess section it looks like it worth processing this case to not to trigger action at all 🤔 

```
    onSuccess: (dispatch, request, response) => {
      if (response === null) return;
      if (request === null) return;
      if (response?.length === 0) return; // new one
      
      dispatch(areaLayersDetailsResourceAtomCache.update(request, response));
    },
```

propakov (approved):
works great! 👍 
found some possible moments to improve, please check

IlyaIzr (approved):

src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsResourceAtom.ts
Akiyamka: Cool!  Extra check removed

PR #415, opened 2023-01-25
fix(core): 13228 - loading message doesn't disappear after using boundary selector and affects overlays

IlyaIzr:
Task has 2 issues so i've did 2 fixes

First fix - boundary selector marker. I've added state for case when no options provided and managed case where first marker stayed on the map due to being managed in async `shedule` callback

Second fix - error was visible when switching 4 times between 2 same layers (i've used bivariate layers). I think it was pretty self-explanatory


src/features/boundary_selector/atoms/boundaryMarkerAtom.ts
```diff
@@ -54,8 +59,13 @@ export const boundaryMarkerAtom = createAtom(
...
 
-    onAction('_setMarker', (marker) => {
-      state = { ...state, marker };
+    onAction('_refreshMarker', ({ marker: newMarker, map }) => {
+      // The only way to update marker content?
+      const previousMarker = state.marker;
+      previousMarker?.remove();
+      newMarker.addTo(map);
+
+      state = { ...state, marker: newMarker };
```
Akiyamka:
```suggestion
      const previousMarker = state.marker;
      state = { ...state, marker: newMarker };
      schedule(() => {
        const previousMarker = state.marker;
        previousMarker?.remove();
        newMarker.addTo(map);
      })
```

src/features/boundary_selector/atoms/boundaryMarkerAtom.ts
```diff
@@ -74,51 +84,58 @@ export const boundaryMarkerAtom = createAtom(
...
               const actions: Action[] = [
-                updateFocusedGeometryAction(selectedFeature),
                 toolbarControlsAtom.disable('BoundarySelector'),
                 updateBoundaryLayerAction(
                   { type: 'FeatureCollection', features: [] },
                   boundaryId,
                 ),
               ];
 
+              if (!selectedFeature) return dispatch(actions);
```
Akiyamka: Why you return dispatch from schedule?
And it seems to me that early return here does more harm than good, it would be easier to read the usual if conditional block

src/features/boundary_selector/atoms/boundaryMarkerAtom.ts
```diff
@@ -74,51 +84,58 @@ export const boundaryMarkerAtom = createAtom(
...
+
+              actions.push(updateFocusedGeometryAction(selectedFeature));
+
               const geometryCamera = getCameraForGeometry(selectedFeature, map);
-              if (typeof geometryCamera === 'object')
+              if (typeof geometryCamera === 'object') {
```
Akiyamka: `typeof null === 'object'`

src/features/layers_in_area/atoms/areaLayersLegendsAndSources.ts
```diff
@@ -84,10 +84,13 @@ export const areaLayersLegendsAndSources = createAtom(
...
+            ? convertDetailsToSource(layerDetails)
+            : null;
+          const existedLayerSource = state.get(layerId)?.data ?? null;
           newState.set(layerId, {
             error: layersDetailsError,
-            data: layerSource,
+            data: layerSourceFromResource || existedLayerSource,
             isLoading: false,
           });
         });
```
Akiyamka: I could not understand the logic of this code.
I understand that we change the behavior here when the backend did not send us the details for requested layer.
So we ask details for layers A, B, C - back-end send data details only for A, C.
Before that in such case we set layer data details to null (sounds logical for me) - now - we trying to keep last data that we get from backend, by set the same last data again? why?

IlyaIzr: @Akiyamka  When we switch between cached layers (previously enabled) - they don't get here
So we have source for layer B in `layersSourcesAtom` but we don't have source for it in resource response

Akiyamka: in that case they not be in requestedLayers

src/features/boundary_selector/atoms/boundaryMarkerAtom.ts
IlyaIzr: okay!
return was used just to end function

src/features/layers_in_area/atoms/areaLayersLegendsAndSources.ts
IlyaIzr: Changed elsewhere

Akiyamka (approved):

PR #422, opened 2023-01-31
feat(areaLayersDetails): 13354 - Calculate Urban Core and Settled Per…

propakov:
…iphery for any focused area

https://kontur.fibery.io/Tasks/Task/Calculate-Urban-Core-and-Settled-Periphery-for-any-focused-area-13354

**Context:** 
Urban Core and Settled Periphery don't recalculate on new drawn geometry. 
But we want to have them recalculated every time we draw something. 
The reason was in cache invalidation. 
After request on api/layers/details we put results in cache. 

Before we had 2 type of record in cache: 
- eventId: Map
- null: Map

Now we extended it with 3rd type:
- hash: true

This is the hash of focusedGeometry, to invalidate layer with boundaryRequiredForRetrieval flag.
So if layer is boundaryRequiredForRetrieval and current focusedGeometry is already in cache - we don't need to request this layer again.

**Notes:**
```
if (layer.eventIdRequiredForRetrieval && !eventId) {
  console.warn(
    `Layer ${layer.id} request is skipped, as eventIdRequiredForRetrieval is true but evenntId is empty.`,
  );
  return false;
}
```
These lines fix a situation when in atom comes eventShape layer, that requires eventId, but eventId is already null.
This layer disappears next update, so this code just prevent invalid request. (check screenshot, 400 error)
Screenshot 2023-01-31 at 11 59 00

### DIFF
--- a/src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
+++ b/src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
@@ -33,7 +33,7 @@ export const areaLayersDetailsParamsAtom = createAtom(
       if (!isEnabled) return false;
 
       if (layer.boundaryRequiredForRetrieval) {
-        const cacheKey: string | undefined = focusedGeometry?.geometry['hash'];
+        const cacheKey = focusedGeometry?.geometry.hash;
         const cached = cacheKey ? cache.get(cacheKey) : false;
         return !cached;
       }
--- a/src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
+++ b/src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
@@ -4,7 +4,10 @@ import { focusedGeometryAtom, getEventId } from '~core/shared_state/focusedGeome
 import { currentEventFeedAtom } from '~core/shared_state';
 import { layersGlobalResource } from '../layersGlobalResource';
 import { layersInAreaAndEventLayerResource } from '../layersInAreaAndEventLayerResource';
-import { areaLayersDetailsResourceAtomCache } from './areaLayersDetailsResourceAtomCache';
+import {
+  areaLayersDetailsResourceAtomCache,
+  getLayersDetailsCacheKey,
+} from './areaLayersDetailsResourceAtomCache';
 import type { DetailsRequestParams } from './types';
 
 export const areaLayersDetailsParamsAtom = createAtom(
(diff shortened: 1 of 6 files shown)


src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsResourceAtomCache.ts
```diff
@@ -3,18 +3,28 @@ import { currentUserAtom } from '~core/shared_state';
...
-/* Store all prev request */
+/* 
+Store all prev request
+Cache structure: 
+  Map 
+  - key: focusedGeometry hash, value: true | For layer with boundaryRequiredForRetrieval=true flag
+  - key: eventId, value: Map | For layer with eventIdRequiredForRetrieval=true flag
+  - key: null, value: Map | For layer with eventIdRequiredForRetrieval=false flag
+*/
+
```
Akiyamka, line 27: nice!

src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
```diff
@@ -30,13 +30,27 @@ export const areaLayersDetailsParamsAtom = createAtom(
 
     const mustBeRequested = allLayers.filter((layer) => {
       const isEnabled = enabledLayers.has(layer.id);
-      if (isEnabled) {
-        const cacheKey = layer.eventIdRequiredForRetrieval ? eventId : null;
-        const cached = cache.get(cacheKey)?.get(layer.id) ?? null;
+      if (!isEnabled) return false;
+
+      if (layer.boundaryRequiredForRetrieval) {
+        const cacheKey: string | undefined = focusedGeometry?.geometry['hash'];
```
Akiyamka: instead of `geometry['hash']` better to align type in
`src/core/shared_state/focusedGeometry.ts`
```ts
export interface FocusedGeometry {
  source: GeometrySource;
  geometry: GeoJSON.GeoJSON & { hash: string };
}
```

src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsResourceAtomCache.ts
```diff
@@ -3,18 +3,28 @@ import { currentUserAtom } from '~core/shared_state';
...
 
-/* Store all prev request */
+/* 
+Store all prev request
+Cache structure: 
+  Map 
+  - key: focusedGeometry hash, value: true | For layer with boundaryRequiredForRetrieval=true flag
+  - key: eventId, value: Map | For layer with eventIdRequiredForRetrieval=true flag
+  - key: null, value: Map | For layer with eventIdRequiredForRetrieval=false flag
+*/
```
Akiyamka: I wonder why only the Boolean flag is stored in the case of geometry? What to do in case new layers are added with the same geometry?

src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
```diff
@@ -30,13 +30,27 @@ export const areaLayersDetailsParamsAtom = createAtom(
...
+        const cached = cacheKey ? cache.get(cacheKey) : false;
```
Akiyamka: what layer you check here?

src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
propakov: I worried it will lead to big refactoring, but it looks much more clearer in the end 👍

src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsResourceAtomCache.ts
propakov: yes, will process this case 
thanks for great case!

src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
propakov: a layer with boundaryRequiredForRetrieval flag

so at first we check if layer is boundaryRequiredForRetrieval
check in cache by hash key

then if no, check if layer eventIdRequiredForRetrieval
if yes - check cache by eventId key
if no - check cache by null key

Akiyamka: "so at first we check if layer is boundaryRequiredForRetrieval" what layer? I expect that you have N such layers

propakov: it's only 2 of them: 
Kontur Urban Core
Kontur Settled Periphery

propakov:
@Akiyamka let's check the last  commit - it applies boundary required layer from cache when you switch event

Akiyamka (approved):

PR #424, opened 2023-01-31
fix(init): 10988 user feeds request no blocking app anymore

Akiyamka:
if this request failed or in takes a lot of time  - app not blocked.
Until we not get available feeds we just not show feed selector.
Bonus - feeds fetched only when feature feed_selector enabled, or when user edit profile

### DIFF
--- a/src/features/events_list/components/FeedSelector/FeedSelector.tsx
+++ b/src/features/events_list/components/FeedSelector/FeedSelector.tsx
@@ -20,7 +20,7 @@ export function FeedSelector() {
     [setCurrentFeed, scheduleAutoSelect],
   );
 
-  return eventFeeds && eventFeeds.length > 1 ? (
+  return eventFeeds?.length > 1 ? (
     <div className={s.feedSelectorContainer}>
       <Text type="short-m">{i18n.t('feed')}:</Text>
       <div>
--- a/src/features/events_list/components/FeedSelector/FeedSelector.tsx
+++ b/src/features/events_list/components/FeedSelector/FeedSelector.tsx
@@ -20,7 +20,11 @@ export function FeedSelector() {
     [setCurrentFeed, scheduleAutoSelect],
   );
 
-  return eventFeeds?.length > 1 ? (
+  if (eventFeeds?.length < 2) {
+    return null;
+  }
+
+  return (
     <div className={s.feedSelectorContainer}>
       <Text type="short-m">{i18n.t('feed')}:</Text>
       <div>
@@ -37,5 +41,5 @@ export function FeedSelector() {
         </select>
       </div>
     </div>
-  ) : null;
+  );
 }
--- a/src/features/events_list/components/FeedSelector/FeedSelector.tsx
+++ b/src/features/events_list/components/FeedSelector/FeedSelector.tsx
@@ -1,20 +1,14 @@
 import { useAction, useAtom } from '@reatom/react';
 import { Text } from '@konturio/ui-kit';
-import { memo, useCallback } from 'react';
+import { useCallback } from 'react';
 import { appConfig } from '~core/app_config';
 import { i18n } from '~core/localization';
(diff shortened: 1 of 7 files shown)


src/features/events_list/components/FeedSelector/index.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { FeedSelector } from './FeedSelector';
+
+export function FeedSelectorFlagged() {
+  const [featureFlags] = useAtom(featureFlagsAtom);
+  const feedSelectorEnabled =
+    featureFlags[FeatureFlag.FEED_SELECTOR] ||
+    featureFlags[FeatureFlag.EVENTS_LIST__FEED_SELECTOR];
+
+  return feedSelectorEnabled && <FeedSelector />;
+}
```
IlyaIzr, line 12: Wouldn't it be better to put this logic into <FeedSelector /> itself? We don't have a feature flag wrapper per-feature elsewhere, and it's expecter from <FeedSelector /> to return `null` if it's not enabled

IlyaIzr:
Wouldn't it break the case when link to non-default feed event will be used?

src/features/events_list/components/FeedSelector/index.tsx
Akiyamka: > Wouldn't it be better to put this logic into itself? We don't have a feature flag wrapper per-feature elsewhere

The main reason I did this is because I can't put a useAtom hook inside an if, that would be a violation of the react hook rule. But I need to do it so I don't have to make an extra request to the api when I don't need to

I also can't check this flag on the top level inside the view - since the feature is nested. 
It seems that we invented widgets to integrate several features, but the only widget I see now doesn't import any of the features (now I'm not sure if we need widgets at all).

As a result, it was decided to make the check as a separate component.

Akiyamka:
> Wouldn't it break the case when link to non-default feed event will be used?

Quite a strange situation that should not normally occur because it breaks the UX, but I do not see what it would break at the code level

src/features/events_list/components/FeedSelector/FeedSelector.tsx
```diff
@@ -26,10 +20,7 @@ const FeedSelectorComp = () => {
     [setCurrentFeed, scheduleAutoSelect],
   );
 
-  return (featureFlags[FeatureFlag.FEED_SELECTOR] ||
-    featureFlags[FeatureFlag.EVENTS_LIST__FEED_SELECTOR]) &&
-    eventFeeds &&
-    eventFeeds.length > 1 ? (
+  return eventFeeds && eventFeeds.length > 1 ? (
```
dqunbp: It is also possible to use optional chaining `eventFeeds?.length > 1`

dqunbp: Another suggestion, return null early

```ts
if(!eventFeeds || eventFeeds <= 1) return null
```

rristoferbowers (approved):

rristoferbowers (approved):

IlyaIzr (approved):

src/features/events_list/components/FeedSelector/FeedSelector.tsx
```diff
@@ -26,10 +20,11 @@ const FeedSelectorComp = () => {
     [setCurrentFeed, scheduleAutoSelect],
   );
 
-  return (featureFlags[FeatureFlag.FEED_SELECTOR] ||
-    featureFlags[FeatureFlag.EVENTS_LIST__FEED_SELECTOR]) &&
-    eventFeeds &&
-    eventFeeds.length > 1 ? (
+  if (eventFeeds?.length < 2) {
```
dqunbp, line 23: To prevent compare `undefined` and 2 we can use 

`eventFeeds?.length ?? 0 < 2`

PR #427, opened 2023-02-02
feat(events): focus event by event bbox

dqunbp:
focus event on click by event geometry bbox
Do not merge until bbox is added to the Event


src/utils/map/cameraForGeometry.ts
```diff
@@ -3,7 +3,7 @@ import { LAPTOP_WIDTH_PX, MOBILE_WIDTH_PX } from '~utils/hooks/useMediaQuery';
 import type { PaddingOptions } from 'maplibre-gl';
 
 // autofocus actual paddings
-function getPaddings(): PaddingOptions {
+export function getPaddings(): PaddingOptions {
   const width = window.visualViewport?.width ?? Infinity;
   // mobile
   if (width < MOBILE_WIDTH_PX + 1) return { left: 64, top: 5, right: 115, bottom: 0 };
```
Komzpa, line 9: please backlog to calculate this from actual grid

dqunbp: Done
https://kontur.fibery.io/Tasks/My-team's-tasks-225#Task/Calculate-event-bbox-paddings-from-actual-grid-14701

Komzpa (approved):

propakov:
types need to be fixed @dqunbp

src/features/events_list/index.tsx
```diff
@@ -2,15 +2,17 @@ import { useAction, useAtom } from '@reatom/react';
 import { useCallback } from 'react';
 import { currentEventAtom } from '~core/shared_state';
 import { scheduledAutoFocus } from '~core/shared_state/currentEvent';
+import { currentEventBbox } from '~core/shared_state/currentEventBbox';
 import { autoSelectEvent } from './atoms/autoSelectEvent';
 import { EventsPanel } from './components/EventsPanel/EventsPanel';
 
 export function EventList() {
   const [currentEvent, currentEventActions] = useAtom(currentEventAtom);
+  const [, bboxActions] = useAtom(currentEventBbox);
```
propakov: maybe useAction would be more suitable?

src/features/events_list/index.tsx
```diff
@@ -2,15 +2,17 @@ import { useAction, useAtom } from '@reatom/react';
...
   const scheduleAutoFocus = useAction(scheduledAutoFocus.setTrue);
```
propakov: scheduleAutoFocus is no more used here, need to remove

src/core/shared_state/currentEventBbox.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    const eventListResource = get('eventListResourceAtom');
+
+    if (!eventListResource || !eventListResource.data) return null;
+
+    const eventData = eventListResource.data.find(
+      (event) => event.eventId === currentEvent.id,
+    );
+
+    if (!eventData) throw new Error('Event not found');
```
propakov: throwing Error without try catch block is bad (the same. for line 32)
it can break reatom, we had a bug like that recently 
it's better to console.warn to tell developers that something happened, or handle them in try/catch block

src/core/shared_state/currentEventBbox.ts
```diff
@@ -0,0 +1,39 @@
+import { createAtom } from '@reatom/core';
+import { eventListResourceAtom } from '~features/events_list/atoms/eventListResource';
```
propakov: content of core shouldn't import stuff from features as far as I remember (because of circular dependencies)

propakov: it looks like this atom should be replaced to ~features/events_list/atoms/

src/features/events_list/index.tsx
```diff
@@ -1,16 +1,16 @@
 import { useAction, useAtom } from '@reatom/react';
 import { useCallback } from 'react';
 import { currentEventAtom } from '~core/shared_state';
-import { scheduledAutoFocus } from '~core/shared_state/currentEvent';
+import { currentEventBbox } from '~features/events_list/atoms/currentEventBbox';
```
propakov: import inside module should be relative  
`import { currentEventBbox } from './atoms/currentEventBbox';`

propakov (approved):
LGTM, just one import to fix. 👍 

Also a question about .husky/pre-push file in commit. Did you do something with it? 
Just see no changes in it and this info in github 100644 → 100755 says nothing to me :)

propakov (commented):
It looks like I've found a bug 
Steps:
- go to next event
- refresh screen 
- you see prev event, but you should see last event you've selected, it looks like you need to update url also

https://user-images.githubusercontent.com/20676525/217843481-c17c64fc-995f-47e8-84c6-655ae4efbd0f.mp4

propakov (changes requested):
need to check/fix this bug before merge

dqunbp:
> need to check/fix this bug before merge

Looks like we don't have the correct BE version on test/dev, `bbox` does not include in event list items

dqunbp:
> It looks like I've found a bug Steps:
> 
> * go to next event
> * refresh screen
> * you see prev event, but you should see last event you've selected, it looks like you need to update url also
> 
>  Feb-09-2023.17-37-57.mp4

@propakov fixed

propakov (approved):
👍

PR #428, opened 2023-02-02
fix(layers_in_area): 14318 - don't update layers in area for /details while loading or on error

IlyaIzr:
fixes error when user had event shape layer turned on and switched to non-event geometry
image

### DIFF
--- a/src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
+++ b/src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
@@ -18,11 +18,9 @@ export const areaLayersDetailsParamsAtom = createAtom(
     state: DetailsRequestParams | null = null,
   ): DetailsRequestParams | null => {
     const layersGlobal = get('layersGlobalResource');
-    const layersInAreaAndEventLayer = get('layersInAreaAndEventLayerResource');
-    const allLayers = [
-      ...(layersGlobal.data ?? []),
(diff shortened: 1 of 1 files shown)


src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
```diff
@@ -18,11 +18,9 @@ export const areaLayersDetailsParamsAtom = createAtom(
     state: DetailsRequestParams | null = null,
   ): DetailsRequestParams | null => {
     const layersGlobal = get('layersGlobalResource');
-    const layersInAreaAndEventLayer = get('layersInAreaAndEventLayerResource');
-    const allLayers = [
-      ...(layersGlobal.data ?? []),
-      ...(layersInAreaAndEventLayer.data ?? []),
-    ];
+    const { loading, error, data } = get('layersInAreaAndEventLayerResource');
+    const layersInAreaAndEventLayer = loading || error || !data ? [] : data;
```
Akiyamka, line 22:
```suggestion
    const layersInAreaAndEventLayer = (loading || error || !data) ? [] : data;
```

Akiyamka (approved):

src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
IlyaIzr: we've skipped it due to lint that keeps fixing that

PR #430, opened 2023-02-02
fix(utils): 14472 createAsyncAtom - incorrect processing error in fetcher

Akiyamka:
Regular throw error in fetcher (not canlcel) of AsyncAtom not reset abortController, and next request was ignored

### DIFF
--- a/src/utils/atoms/createAsyncAtom/createAsyncAtom.ts
+++ b/src/utils/atoms/createAsyncAtom/createAsyncAtom.ts
@@ -169,7 +169,7 @@ export function createAsyncAtom<
               log('5.1.B.1. Error');
               dispatch(create('_error', params, ABORT_ERROR_MESSAGE));
             } else if (ctx.abortController === abortController) {
-              ctx.abortController = null;
+              delete ctx.abortController;
               log('5.1.C Not abort error');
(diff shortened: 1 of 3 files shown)


src/utils/atoms/createAsyncAtom/createAsyncAtom.ts
```diff
@@ -165,6 +169,8 @@ export function createAsyncAtom<
               log('5.1.B.1. Error');
               dispatch(create('_error', params, ABORT_ERROR_MESSAGE));
             } else if (ctx.abortController === abortController) {
+              ctx.abortController = null;
```
dqunbp: `delete ctx.abortController;` will do the same as `ctx.abortController = null;`?

Akiyamka:
```suggestion
              delete ctx.abortController;
```

Akiyamka: yep

propakov: let's do it in one approach. 
If we go with delete, could you also please update in line 201?

propakov (approved):

PR #431, opened 2023-02-03
fix(layers): add appId to details request

vkozio:


IlyaIzr:
we need to merge to release branch

PR #432, opened 2023-02-06
feat(doc): 14682 - improved doc about translations by adding step by …

propakov:
…step guide

https://kontur.fibery.io/Tasks/Task/Improve-doc-about-translations-by-adding-step-by-step-guide-14682


README.md
```diff
@@ -38,20 +38,78 @@ npm run dev
...
+    <a
+        href="https://www.kontur.io/portfolio/event-feed/"
+        target="_blank"
+        rel="noreferrer"
+    >
+        Kontur Event Feed
+    </a>
+    Some content after
+</Trans>
+```
```
Akiyamka, line 100: Maybe we should add that key difference between `i18n.t()` and `<Trans />` - last one track current language and change translation right after language changed

Akiyamka (approved):
lgfm

README.md
```diff
@@ -38,20 +38,78 @@ npm run dev
...
+3. Run update script:
+```
+npm run i18n:update
+```
+And then fill translations for extracted keys to src/core/localization/translations/en/common.json.
+If you use ```<Trans/>``` translation will be extracted automatically: ```"p4": "Some content before<1>Kontur Event Feed</1>Some content after"```
```
IlyaIzr: what do you mean by that?
I thought after running the command all new keys will be extracted

README.md
```diff
@@ -38,20 +38,78 @@ npm run dev
...
 
-We use 18next + gettext convertion approach in our development toolchain.
-There are several tasks in package.json:
 - i18n:export converts i18next to gettext
 - i18n:import converts gettext to i18next
-- i18n:update extracts new translation keys from code to common.json (i18n.t, <Trans/>) and removes unused keys.
-- lint:i18n:keys:identity is needed to see untranslated keys for all locales (make sure you ran i18n:import before, it compares i18n files)
+- i18n:update extracts new translation keys from code to common.json (```i18n.t``` or ```<Trans/>```) and also removes unused keys
+- lint:i18n:keys:identity is needed to see untranslated keys for all locales (make sure you ran i18n:import before as it compares i18next .json files)
 - i18n:msgmerge is needed to keep all .po files synced with .pot file (marks non-compliant msgid as 'fuzzy', renames keys in .po according keys changes in .pot)
```
IlyaIzr: very unobvious command name btw, but as long as we have doc - it's fine
(i'd expect smth like "updatepot" or 'syncpot' or in js world i'd go with "updateTranslationsArchive" but that's too long for those who type in console :D)

IlyaIzr (approved):

README.md
propakov: Yep, that's right. But ```<Trans/>``` has also translation inside:
```
<Trans i18nKey="about.p4">
    Some content before
    <a
        href="https://www.kontur.io/portfolio/event-feed/"
        target="_blank"
        rel="noreferrer"
    >
        Kontur Event Feed
    </a>
    Some content after
</Trans>
```
So it will extract key + translation.

README.md
propakov: hm, msgmerge is a name of gettext package util, so that's why  🤷‍♂️ 
maybe i18n:gettext:sync ?

README.md
propakov: added to the 1. point

README.md
IlyaIzr: So it's like that maybe?
```suggestion
If you use ```<Trans/>``` translations from it will be extracted automatically: ```"p4": "Some content before<1>Kontur Event Feed</1>Some content after"```
```

README.md
IlyaIzr: looks better too

README.md
propakov: sure if it's clearer 👍

PR #434, opened 2023-02-06
fix(current_event): 12434 - event auto update reset focused geometry when episode selected

Akiyamka:
Episode and event can be selected at the same time, in which case the geometry of the episode is more important, and we not override it


propakov (commented):
Is it a bug? Works differently on main. 
Steps: 
- select event
- select episode from timeline
- see geometry visible
- select different event
result: no geometry
expected result: new event geometry

https://user-images.githubusercontent.com/20676525/217454551-91d3ef2c-5592-4269-939d-92dd70cd0787.mp4

dqunbp (approved):

propakov:
@Akiyamka doing the same steps from previous video, but not every time I've reproduce next error:
Screenshot 2023-02-10 at 09 31 18
Screenshot 2023-02-10 at 09 30 58

propakov (approved):
works like a charm  👍

PR #439, opened 2023-02-09
feat(live_sensor): 14263 - create live sensor feature

IlyaIzr:
Here we introduce live sensor feature. It's somewhat between POC and MVP
The goal of this feature is to collect sensor data on mobile phone and send it in geojson feature format.

It's intended to be used on Smart City app only so far. For logged in users that have sensors on their device. 
Readme attached!

[See diagram in figma](https://www.figma.com/file/GRMz4BnDfr5qFXafmrMt9Y/Live-sensor-feature?node-id=0%3A1&t=w2FGK3oikxVbA5QU-1)


src/features/live_sensor/atoms/sensorResource.ts
```diff
@@ -0,0 +1,47 @@
+import { apiClient } from '~core/apiClientInstance';
+import { createAtom } from '~utils/atoms';
+import { createAsyncAtom } from '~utils/atoms/createAsyncAtom';
+import { createNumberAtom } from '~utils/atoms/createPrimitives';
+import { UPDATE_ENDPOINT_PATH } from '../constants';
+import { collectedPointsAtom } from './collectedPoints';
+import type { FeatureCollection } from '~utils/geoJSON/helpers';
+
+// Here we want to send data every each REQUESTS_INTERVAL
```
Komzpa: We want to send once per gps point. The data without new gps point is not helping.

src/features/live_sensor/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+          title: "Your device don't have needed sensors",
+        });
+      }
+      if (featureInitializingFailed) return;
+
+      if (!isActive) {
+        this.title = 'Start sensor recording';
+        stopRecording();
+        notificationServiceInstance.info({
+          title: 'Recording has been finished',
```
Komzpa: i18n

src/features/live_sensor/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+    icon: <Car24 />,
+    onClick: (isActive) => {
+      if (isActive) {
+        toolbarControlsAtom.enable.dispatch(SENSOR_CONTROL);
+      } else toolbarControlsAtom.disable.dispatch(SENSOR_CONTROL);
+    },
+    onChange(isActive) {
+      if (featureInitializingFailed && isActive) {
+        notificationServiceInstance.error({
+          title: "Your device don't have needed sensors",
```
Komzpa: i18n

src/features/live_sensor/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export function getOnErrorFunction(
+  notificationServiceInstance: NotificationService,
+  stopRecording: () => void,
+) {
+  return function onError(event: SensorErrorEvent | GeolocationPositionError) {
+    const eventName = 'code' in event ? event.code + event.message : event.error.name;
+    const eventMessage = 'message' in event ? event.message : event.error.message;
+    notificationServiceInstance.warning({
+      title: eventName || "Can't connect to server",
```
Komzpa: i18n

src/features/live_sensor/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+  // It calls prompt window to allow sharing location for the first time
+  const watchId = geolocation.watchPosition((pos) => {
+    // This function runs each second after user allowed sharing navigation
+    sensorDataAtom.updateSensor.dispatch('coordinates', {
+      lng: pos.coords.longitude,
+      lat: pos.coords.longitude,
+      alt: pos.coords.altitude,
+      accuracy: pos.coords.accuracy,
+      speed: pos.coords.speed,
+      course: pos.coords.heading,
```
Komzpa: why did you rename heading to course here? heading is pretty standard gps term.

src/features/live_sensor/constants.ts
```diff
@@ -0,0 +1,11 @@
+export const SENSOR_CONTROL = 'sensorControl';
+export const SENSOR_CONTROL_NAME = 'Live sensor control';
+
+export const UPDATES_PER_MINUTE = 60;
+// in milliseconds
+export const REQUESTS_INTERVAL = 1000;
+export const ADDING_POINTS_INTERVAL = 100;
```
Komzpa: What is this

src/features/live_sensor/atoms/collectedPoints.ts
```diff
@@ -0,0 +1,41 @@
...
+  {
+    addFeature: () => null,
+    reset: () => null,
+  },
+  ({ onAction, getUnlistedState }, state: FeatureCollection = defaultState) => {
+    const sensorData = getUnlistedState(sensorDataAtom);
+
+    onAction('addFeature', () => {
+      if (!sensorData) return state;
+      const { coordinates, ...otherProps } = sensorData.data;
```
Komzpa: One point per accelerometer reading is a no-go.

Komzpa (changes requested):
Needs pa bit more work

src/features/live_sensor/atoms/collectedPoints.ts
IlyaIzr: what made you think it creates point on accelerometer reading?

src/features/live_sensor/constants.ts
IlyaIzr: interval of creating geojson points that will come in array to send to back end

src/features/live_sensor/utils.ts
IlyaIzr: okay, i'll go standard way
Did it because our source example had "heading"

src/features/live_sensor/atoms/sensorResource.ts
IlyaIzr: if gps updates once per second - we might miss some spikes in that second interval
let's discuss that on the meeting

IlyaIzr:
i will redo logic a bit to hook updates more on gps update event

IlyaIzr:
taking care of localization error now...

src/core/localization/translations/en/common.json
```diff
@@ -292,5 +292,12 @@
     "body": "We use absolutely necessary cookies to provide you personalized services and optional cookies to improve Disaster Ninja and your experience. You can manage cookies settings or withdraw consent to optional cookies at any time.\nFor more information, please, check our [Privacy Policy](about/privacy)",
     "decline_all": "Decline optional cookies",
     "accept_all": "Accept optional cookies"
+  },
+  "live_sensor": {
+    "start": "Start sensor recording",
+    "finish": "Stop sensor recording",
+    "finishMessage": "Recording has been finished",
+    "startMessage": "Recording has been started",
+    "noSensorsError": "Your device don't have needed sensors"
```
Komzpa:
```suggestion
    "noSensorsError": "Your device desn't have the required sensors"
```

src/features/live_sensor/atoms/sensorData.ts
```diff
@@ -0,0 +1,70 @@
...
+      } else if (id === 'gyroscope') {
+        newState.data.gyroscope = data as Gyroscope;
+      } else if (id === 'accelerometer') {
+        newState.data.accelerometer = data as Accelerometer;
+      } else if (id === 'coordinates') {
+        newState.data.coordinates = data as Coordinates;
+      }
+
+      newState.timestamp = new Date().getTime();
+      state = newState;
```
Komzpa, line 67: I see you are replacing state with new state. You need to collect it to the array instead. Or where is the array composition happening?

src/core/localization/gettext/ar/common.po
```diff
@@ -8,491 +8,491 @@ msgstr ""
...
 msgstr "الاتجاه الأفقي"
 
-#: legend_presentation
+#: legend_presentation
 msgid "Legend presentation"
 msgstr "عرض العنوان التفسيري"
 
-#: layers bivariate##color_manager##layers_filter
+#: layers
 bivariate##color_manager##layers_filter
```
propakov: hm, it's not ok. 
how it happened? after i18n:gettext-sync ? 
so after I run this command on your branch - result is next: 
Screenshot 2023-02-10 at 18 11 17

propakov: what is your msgmerge version? 
msgmerge --version
msgmerge (GNU gettext-tools) 0.21

propakov: I've pushed results of command run to unblock you
Need to investigate why it happened

IlyaIzr: same version as yours
image

I think it was from sync, lesser possibility - from update or export commands for i18n

src/features/live_sensor/atoms/sensorData.ts
IlyaIzr: weird, I've changed this logic on [429ffc5](https://github.com/konturio/disaster-ninja-fe/pull/439/commits/429ffc5620040158d74088175eb28a6446388815)

Komzpa: I don't see the change and I'm exactly in the commit you mentioned :)

src/features/live_sensor/constants.ts
```diff
@@ -0,0 +1,10 @@
+export const SENSOR_CONTROL = 'sensorControl';
+export const SENSOR_CONTROL_NAME = 'Live sensor control';
+
+export const UPDATES_PER_MINUTE = 60;
+// in milliseconds
+export const RECORD_UPDATES_INTERVAL = 100;
```
Komzpa: both const seem not needed if you push to server based on location update

src/features/live_sensor/utils.ts
```diff
@@ -0,0 +1,90 @@
...
+    // This function runs each second after user allowed sharing navigation
+    // Run updating sensor first
+    sensorDataAtom.updateSensor.dispatch('coordinates', {
+      lng: pos.coords.longitude,
+      lat: pos.coords.longitude,
+      alt: pos.coords.altitude,
+      accuracy: pos.coords.accuracy,
+      speed: pos.coords.speed,
+      heading: pos.coords.heading,
+      timestamp: pos.timestamp,
```
Komzpa: for investigations please report the timestamp in pos object and the current time on the system clock. in case of GPS they are expected to be different, in case of wifi - very close, which might help filter out some updates later. cc @frolui

src/features/live_sensor/utils.ts
```diff
@@ -0,0 +1,90 @@
...
+  }, getOnErrorFunction(stopRecording, 'geolocation'));
+  return watchId;
+}
+
+function getOnErrorFunction(stopRecording?: () => void, sensorName?: string) {
+  return function onError(event: SensorErrorEvent | GeolocationPositionError) {
+    const eventName = 'code' in event ? event.code + event.message : event.error.name;
+    const eventMessage = 'message' in event ? event.message : event.error.message;
+    notificationServiceInstance.warning({
+      title: eventName || `Can't connect to sensor ${sensorName}`,
```
Komzpa: i18n

src/core/localization/gettext/es/common.po
```diff
@@ -956,6 +956,26 @@ msgstr ""
...
+#: live_sensor##finishMessage
+msgid "Recording has been finished"
+msgstr ""
+
+#: live_sensor##startMessage
+msgid "Recording has been started"
+msgstr ""
+
+#: live_sensor##noSensorsError
+msgid "Your device don't have needed sensors"
```
Komzpa: rebuild i18n?

src/features/live_sensor/constants.ts
IlyaIzr: `RECORD_UPDATES_INTERVAL ` defines how often we add data to array before sending it

src/features/live_sensor/atoms/sensorData.ts
IlyaIzr: github bug?
this code is 100% deleted on the repo

src/features/live_sensor/utils.ts
IlyaIzr: added, not resolving in case @frolui didn't seen it yet

src/features/live_sensor/constants.ts
IlyaIzr: UPDATES_PER_MINUTE was obsolete and deleted, yes, thank you!

IlyaIzr:
[Added diagram in figma](https://www.figma.com/file/GRMz4BnDfr5qFXafmrMt9Y/Live-sensor-feature?node-id=0%3A1&t=w2FGK3oikxVbA5QU-1)

src/features/live_sensor/readme.md
```diff
@@ -12,10 +12,55 @@ Meanwhile `collectedPointsAtom` will collect geojson points created from sensor
...
+  features: [
+    // point created on first 100ms (if RECORD_UPDATES_INTERVAL === 100)
+    {
+      type: 'Feature',
+      geometry: {
+        type: 'Point',
+        coordinates: [coordinates.lng, coordinates.lat],
+      },
+      properties: {
+        accelerometer: { ...accelerometerSensorData },
```
Komzpa: this should be array of all the accelerometer observations in last 1 second, not just one accelerometer observation

src/features/live_sensor/readme.md
```diff
@@ -12,10 +12,55 @@ Meanwhile `collectedPointsAtom` will collect geojson points created from sensor
...
+
+#### Output
+
+For the request we'll sent following JSON:
+
+```js
+{
+  type: 'FeatureCollection',
+  features: [
+    // point created on first 100ms (if RECORD_UPDATES_INTERVAL === 100)
```
Komzpa, line 27: this should be created not based on timer, but based on event (gps location acquired).

one gps coordinate event should be used only in one message to the server

Komzpa (changes requested):

src/features/live_sensor/constants.ts
```diff
@@ -0,0 +1,9 @@
+export const SENSOR_CONTROL = 'sensorControl';
+export const SENSOR_CONTROL_NAME = 'Live sensor control';
+
+// in milliseconds
+export const RECORD_UPDATES_INTERVAL = 100;
+
+export const UPDATE_ENDPOINT_PATH = '/features/live-sensor';
+
+export const SENSOR_PRESICION = 6;
```
vkozio: typo: PRECISION

src/features/live_sensor/readme.md
IlyaIzr: yup, discussing new architecture ATM

IlyaIzr:
discussing new architecture ATM ❗ 
changes - hardcoded interval of updates is removed
sensors data won't be stored per GeoJson point but as a array in single point per location update

src/features/live_sensor/atoms/collectedPoints.ts
```diff
@@ -0,0 +1,43 @@
...
+      state.features = [...state.features, feature];
+    });
+
+    onAction('reset', () => {
+      state = defaultState;
+    });
+
+    return state;
+  },
+  'collected live sensor points atom',
```
propakov: What's our flow about atom id? I thought it's named like atom name, so here I would expect collectedPointsAtom. 
Or it's doesn't matter?

src/features/live_sensor/atoms/sensorData.ts
```diff
@@ -0,0 +1,73 @@
+import { createAtom } from '~utils/atoms';
+
+type uncertainNumber = number | null;
```
propakov: Type always first letter capital  -> UncertainNumber. Or there is some reason why it's lower case?

propakov:
tested it a bit
I have denied geolocation
I click start recording button, it calls stopRecording(), that stops all sensors
it's fine, but button still available and recording goes on 

comparing with "Locate me", I would expect that button will be deactivated and recording stops
or is it fine for now? 
Screenshot 2023-02-13 at 17 55 52

src/features/live_sensor/atoms/collectedPoints.ts
IlyaIzr: i dunno
will do

propakov: just checked other atoms naming 🤷‍♂️ 
let's keep it consistent

IlyaIzr:
@propakov considering behavior - it's okay we don't have any proper user flow planned so far

src/features/live_sensor/readme.md
```diff
@@ -0,0 +1,63 @@
...
+,
+        orientX: {1, .9, .8, .9},
+        orientY: {1, .9, .8, .9},
+        orientZ: {1, .9, .8, .9},
+        orientW: {1, .9, .8, .9},
+,
+        gyroX: {1, .9, null, null},
+        gyroY: {1, .9, null, null},
+        gyroZ: {1, .9, null, null},
+        lng: {22.11, 22.1155, 22.11556, 22.11531},
```
Komzpa: what? where does multiple gps points come from?

IlyaIzr: yeah
my bad
wrong example!

IlyaIzr: will fix another error that this is object instead of array

src/features/live_sensor/atoms/sensorData.ts
```diff
@@ -0,0 +1,144 @@
...
+};
+type Gyroscope = {
+  gyroX: UncertainNumber;
+  gyroY: UncertainNumber;
+  gyroZ: UncertainNumber;
+  timestamp: number;
+};
+
+// every value refers to updatedAt by it's index
+// accelY[55] was set at updatedAt[55]
```
propakov: can't see updatedAt in PR, what is it about?

src/features/live_sensor/utils.ts
```diff
@@ -0,0 +1,97 @@
...
+    // This function runs each second or more after user allowed sharing navigation
+    collectedPointsAtom.addFeature.dispatch({
+      lng: pos.coords.longitude,
+      lat: pos.coords.longitude,
+      alt: pos.coords.altitude,
+      accuracy: pos.coords.accuracy,
+      speed: pos.coords.speed,
+      heading: pos.coords.heading,
+      coordTimestamp: pos.timestamp,
+      coordSystTimestamp: new Date().getTime(),
```
propakov: getTime() without new Date().getTime(),

IlyaIzr:
need to simlify a bit

IlyaIzr:
implemented cahnges

Komzpa (approved):
jsons LGTM

src/features/live_sensor/atoms/sensorData.ts
```diff
@@ -0,0 +1,129 @@
...
+        if (sensorData && outdatedDataIndex) {
+          sensorData = sensorData.slice(outdatedDataIndex + 1);
+        }
+        if (sensorData) sensorData.push(data[sensorKey]);
+
+        newState[sensorKey] = sensorData || [data[sensorKey]];
+      });
+
+      state = newState;
+    }
```
Akiyamka: I propose to take this function out of the reducer, and add test for it.

Akiyamka: recreating the function on every update looks like a waste of resources here

src/features/live_sensor/atoms/sensorResource.ts
```diff
@@ -0,0 +1,55 @@
...
+    const newState = getUnlistedState(collectedPointsAtom);
+    state = { ...newState };
+
+    if (state?.features.length) {
+      // schedule an effect to clear collected data after request was triggered
+      schedule((dispatch) => dispatch(sensorDataAtom.resetAllData()));
+    }
+    return state;
+  },
+);
```
Akiyamka: it seems to have nothing to do with resource dependencies, better to make a separate atom for this
```
 if (state?.features.length) {
      // schedule an effect to clear collected data after request was triggered
      schedule((dispatch) => dispatch(sensorDataAtom.resetAllData()));
    }
 ```
 
 In general you don't need a dependency atom here, you can pass null instead, and call the .request action in the required queries manually

File: src/features/live_sensor/atoms/sensorResource.ts [hunk 25]
```diff
@@ -0,0 +1,55 @@
...
+
+      // reset on successful request
+      collectedPointsAtom.reset.dispatch();
+      return res;
+    } catch (error) {
+      console.warn(error);
+      // don't reset on error (keep collection)
+    }
+  },
+  'Sensor_resource_atom',
```
Akiyamka:
```suggestion
  'sensorResourceAtom',
```

src/features/live_sensor/constants.ts
```diff
@@ -0,0 +1,9 @@
+export const SENSOR_CONTROL = 'sensorControl';
+export const SENSOR_CONTROL_NAME = 'Live sensor control';
+
+export const UPDATE_ENDPOINT_PATH = '/features/live-sensor';
```
Akiyamka: Usually we write the path directly in the query, why put it in a constant here?

src/features/live_sensor/atoms/sensorResource.ts
```diff
@@ -0,0 +1,55 @@
...
+          ...sensorFeatures,
+        },
+        true,
+        {
+          signal: abortController.signal,
+        },
+      );
+
+      // reset on successful request
+      collectedPointsAtom.reset.dispatch();
```
Akiyamka: resource only receives data, this is already a side effect, you should not do so

src/features/live_sensor/atoms/sensorResource.ts
IlyaIzr: still need to trigger reset only after request ran

src/features/live_sensor/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+    name: SENSOR_CONTROL_NAME,
+    title: i18n.t('live_sensor.start'),
+    active: false,
+    visualGroup: controlVisualGroup.noAnalytics,
+    exclusiveGroup: controlGroup.mapTools,
+    icon: <Car24 />,
+    onClick: (isActive) => {
+      if (isActive) {
+        toolbarControlsAtom.enable.dispatch(SENSOR_CONTROL);
+      } else toolbarControlsAtom.disable.dispatch(SENSOR_CONTROL);
```
Akiyamka:
```
if (isActive) {
   toolbarControlsAtom.enable.dispatch(SENSOR_CONTROL);
} else {
   toolbarControlsAtom.disable.dispatch(SENSOR_CONTROL);
} 
```

src/features/live_sensor/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+        notificationServiceInstance.info({
+          title: i18n.t('live_sensor.finishMessage'),
+        });
+        return;
+      }
+
+      this.title = i18n.t('live_sensor.finish');
+      resourceUnsubscribe = sensorResourceAtom.subscribe(() => {
+        /*noop*/
+      });
```
Akiyamka: not needed if you trigger resource manually

src/features/live_sensor/utils.ts
```diff
@@ -0,0 +1,105 @@
...
+import type { TriggerRequestActionType } from './atoms/triggerResource';
+
+export type UncertainNumber = number | null;
+
+export function hookSensors(
+  sensorDataAtom: SensorDataAtomExportType,
+  stopRecording: () => void,
+  accelerometer: Accelerometer,
+  orientationSensor: AbsoluteOrientationSensor,
+  gyroscope: Gyroscope,
```
Akiyamka, line 13: If you have more than 3 arguments it's better to use named arguments
function withManyArgs({ foo, bar baz })

Akiyamka (changes requested):
1. Remove side effects from fetcher.
2. Directly .request resource instead of hacks with triggerResource
3. Move update function declaration from reducer

In general the code lacks structure, looks confusing, needs refactoring before adding new features here

src/features/live_sensor/atoms/collectedPoints.ts
```diff
@@ -0,0 +1,65 @@
...
+export const collectedPointsAtom = createAtom(
+  {
+    addFeature: (data: CoordinatesData) => data,
+    reset: () => null,
+  },
+  ({ onAction, getUnlistedState, schedule }, state: FeatureCollection = defaultState) => {
+    const sensorData = getUnlistedState(sensorDataAtom);
+
+    onAction('addFeature', (data) => {
+      if (!sensorData) return state;
```
dqunbp, line 33: Do we need to add the feature `data` if exists to `state`?

IlyaIzr: sorry didn't understand the question, can you be more verbose?

src/features/live_sensor/atoms/sensorData.ts
```diff
@@ -0,0 +1,129 @@
...
+      if (!incomingTime) {
+        console.error('no timestamp provided', data);
+        return;
+      }
+
+      let outdatedDataIndex: number | null = null;
+
+      if (sensorTimes?.length) {
+        // Find if older parts of array are older than SENSOR_DATA_LIFETIME
+        for (let i = 0; i < sensorTimes.length; i++) {
```
dqunbp: Is it more efficient to start from the end of the array and break on the first outdated item?

IlyaIzr: the order is opposite
`(older values) [1000, 1001, 1002, 1003, 1004] (newer values)`

src/features/live_sensor/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+  try {
+    accelerometer = new Accelerometer();
+    orientationSensor = new AbsoluteOrientationSensor();
+    gyroscope = new Gyroscope();
+    geolocation = navigator.geolocation;
+  } catch (e) {
+    featureInitializingFailed = true;
+  }
+
+  let interval: NodeJS.Timer;
```
dqunbp: Is there a place where the "interval" value is set?

IlyaIzr: hehe i forgot to remove it, thanks

src/features/live_sensor/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+        this.title = i18n.t('live_sensor.start');
+        stopRecording();
+        notificationServiceInstance.info({
+          title: i18n.t('live_sensor.finishMessage'),
+        });
+        return;
+      }
+
+      this.title = i18n.t('live_sensor.finish');
+      atomUnsubscribe = resourceWatcherAtom.subscribe(() => {
```
Akiyamka: unsubscribeWatcherAtom

src/features/live_sensor/atoms/collectedPoints.ts
```diff
@@ -0,0 +1,73 @@
...
+          ...sensorData,
+          ...data,
+          userAgent: navigator.userAgent,
+        },
+      };
+
+      state.features = [...state.features, feature];
+
+      // Trigger request after each new feature was added
+      // schedule((dispatch) => dispatch(resourceWatcherAtom.triggerRequest()));
```
Akiyamka: leftover of the old code

src/features/live_sensor/atoms/sensorResource.ts
```diff
@@ -0,0 +1,28 @@
...
+  null,
+  async (sensorFeatures, abortController) => {
+    if (!sensorFeatures.features.length) return null;
+
+    try {
+      await apiClient.post(
+        '/features/live-sensor',
+        {
+          ...sensorFeatures,
+        },
```
Akiyamka: maybe extra copy, api not mutate payload by any way.

Akiyamka (approved):
All requested changes added

PR #444, opened 2023-02-13
Set first list event index to selected event index

dqunbp:


src/features/events_list/components/FullState/FullState.tsx
```diff
@@ -28,33 +28,38 @@ export function FullState({
   const hasTimeline = featureFlags[FeatureFlag.EPISODES_TIMELINE];
   const virtuoso = useRef<VirtuosoHandle>(null);
 
+  const [currentEventIndex, setCurrentEventIndex] = useState<number | null>(() =>
```
propakov: it seems like you don't need a null here, because in falsy case it's 0

if you would introduce function to get eventIndex like 
```
  const getCurrentEventIndex = (
    eventsList: Event[] | null,
    currentEventId?: string | null,
  ): number => {
    if (!currentEventId || !eventsList?.length) return 0;

    const currentEventIndex = eventsList.findIndex(
      (event) => event.eventId === currentEventId,
    );

    return currentEventIndex > -1 ? currentEventIndex : 0;
  };
```
you could use it for state init
```
  const [currentEventIndex, setCurrentEventIndex] = useState<number>(() =>
    getCurrentEventIndex(eventsList, currentEventId),
  );
```
and also in   useLayoutEffect
and no need to handle falsy in component prop here as it's always a number 
`initialTopMostItemIndex={currentEventIndex}`

propakov:
while testing find a bug 
1 select any event in event list
2 refresh page
3 event is not visually selected

Screenshot 2023-02-14 at 10 52 00

dqunbp:
> while testing find a bug 1 select any event in event list 2 refresh page 3 event is not visually selected
> 
> Screenshot 2023-02-14 at 10 52 00

Thanks for testing, fixed

src/features/events_list/components/FullState/FullState.tsx
dqunbp: @propakov `null` required here
I can describe how it works:
1. `currentEventIndex` is `null`, until it will be found in `event_list`
2. `event_list` loading will be true until `currentEventIndex` is null
3. When `currentEventIndex` is founded or not it will become a number
4. When the `currentEventIndex` is a number `event_list` will be rendered with the valid value of `currentEventIndex`

Otherwise, if `currentEventIndex` always will be number, we always will get `initialTopMostItemIndex=0` at the first `event_list` render

propakov: thanks for explanation! it became more clear

not clear why loading waits for currentEventIndex, if eventsList.findIndex is not an async operation

propakov: it feels like it all can be simplified, try to get rid of all useEffects and useState next way 

```
  const currentEventIndex = useMemo(() => {
    if (!currentEventId && !loading && eventsList?.length) return 0;
    if (!currentEventId || !eventsList?.length) return null;

    const index = eventsList.findIndex((event) => event.eventId === currentEventId);

    if (index > -1) {
      return index;
    } else {
      return 0;
    }
  }, [currentEventId, eventsList, loading]);
  
   const hasUnlistedEvent = currentEventIndex && currentEventIndex > 1;
```

propakov: @Akiyamka what do you think on this? 👀

dqunbp: > not clear why loading waits for currentEventIndex, if eventsList.findIndex is not an async operation

But render is async

Unfortunately, this solution does not work correctly with `hasUnlistedEvent`

We lost the information, is `currentEventIndex` found or not

It should stay `false` if `currentEventIndex` was not found, but here, during the loading, we will have `false` initially and then `true` if the selected event has an index > 0, which is not correct.

If `currentEventIndex` is found and equals 0, it will be `false` which is not correct as well because it should be `true` in case `currentEventIndex` is not found in `eventList` only

Akiyamka: I think we can express this logic more clearly and transparently, and, following the principle of sole responsibility, divide this logic into two parts

```suggestion

const statesToComponents = createStateMap({
    error,
    loading: loading,
    data: eventsList,
  });

  const currentEventIndex = useMemo(() => {
    if (!currentEventId) return 0;
    if (eventsList === null || eventsList.length === 0) return 0;

    const index = eventsList.findIndex((event) => event.eventId === currentEventId);
    return Math.max(index, 0);
  }, [currentEventId, eventsList]);

  const eventListIncludesCurrentEvent: true | false | 'unknown' = useMemo(() => {
    if (error || loading || eventsList === null) return 'unknown';
    return !!eventsList.find((event) => event.eventId === currentEventId);
  }, [currentEventId, eventsList, error, loading]);

<...>

  return (
    <div className={s.panelBody}>
      {eventListIncludesCurrentEvent === false && (
        <CurrentEvent hasTimeline={Boolean(hasTimeline)} showDescription={true} />
      )}
```

Akiyamka (approved):

PR #445, opened 2023-02-13
fix(core): 13184 update vite and vitest fix

vkozio:
https://kontur.fibery.io/Tasks/Task/Update-vite-and-vitest-13184

update to vite 4.1, vitest 0.28.5
vite config now uses path resolutions from ts config 
treeshaking fixes
lodash to lodash-es (esm), converted some lodash to native. lodash has issues with treeshaking, we should move away from it
some deps updates
CI: test runner uses test env, don't run tests twice


IlyaIzr:
Why converting only some lodash methods? Looks like semi-measure that solves nothing at a first glance

IlyaIzr:
Overall looks okay BUT seems like this could've been 2 or 3 PR's
Why bother? 
While you're working on CI issues, simplier refactors could've been solved and merged. Also would be easier to read

can't approve at this stage while CI fails for uknown reason

vkozio:
> Why converting only some lodash methods? Looks like semi-measure that solves nothing at a first glance

it fixed issue with rollup bundling entire lodash, so it's ok for now
remaining methods can be replaced in scope of other tasks

.github/workflows/build.yml
```diff
@@ -48,8 +48,7 @@ jobs:
       - run: npm run postinstall
       - run: npm run lint
       - run: npm run typecheck
-      - run: npm run test:unit
-      - run: npm run coverage
+      - run: env NODE_ENV=test npm run coverage
```
IlyaIzr, line 51: `npm run coverage` starts test?

what's not obvious - 1. removing `test:unit` from CI
2. set NODE_ENV specifically ==='test'

IlyaIzr: discussed "coverage" question

vkozio: test supposed to run with NODE_ENV = 'test' but new vitest pick value from environment (production) and react loads production bundle - without test utils like act() - which produces error

IlyaIzr (approved):

PR #446, opened 2023-02-14
feat(profile): 14465 - every language option is translated on it's ow…

propakov:
…n locale

https://kontur.fibery.io/Tasks/Task/The-names-of-the-languages-translated-into-the-selected-language-14465
Screenshot 2023-02-15 at 14 55 40
Screenshot 2023-02-15 at 14 54 52
Screenshot 2023-02-15 at 14 54 28


src/core/localization/gettext/de/common.po
```diff
@@ -923,6 +899,11 @@ msgstr "Standardmäßiger OpenStreetMap-Editor (Beta)"
 msgid "All changes have been applied successfully"
 msgstr "Alle Änderungen wurden erfolgreich übernommen"
 
+# it's a current locale's language - German, pls ignore msgid
+#: profile##languageOption
+msgid "English"
+msgstr "Deutsch"
```
dqunbp, line 926: This approach looks strange for translators)

src/core/localization/gettext/ko/common.po
```diff
@@ -854,30 +854,6 @@ msgstr "영국식 단위(베타)"
...
-msgid "Spanish"
-msgstr "스페인어"
-
-#: profile##arabicLanguageOption
-msgid "Arabic"
-msgstr "아랍어"
-
-#: profile##koreanLanguageOption
-msgid "Korean"
-msgstr ""
```
dqunbp, line 871: Is it an option to write each language name as is, except for the file target language?

E.g.

```po
msgid "English"
msgstr "English"

#: profile##spanishLanguageOption
msgid "Spanish"
msgstr "Español"

msgid "Korean"
msgstr "한국어"
```

src/core/localization/gettext/de/common.po
propakov: what do you suggest? 
bearing in mind that we have i18n:gettext-sync, that merge by msgid

translators will never touch it, as it's already translated.

src/core/localization/gettext/ko/common.po
propakov: because of i18n:gettext-sync, we sync keys between .po files
so if you have 

#: profile##spanishLanguageOption
msgid "Spanish"
msgstr "Español"

in spanish .po files, you need to have it in all other files too

dqunbp: Yes, we can have the same translations like this in all `po` files, will it work?

propakov: hm, i didn't think about it, it will 
so idea is to have 6 repeating keys in all files on different languages
it's also confusing for translator for sure, but maybe better
maybe better to ask for other devs opinion on frontend channel

propakov: reimplemented on your approach  🚀

dqunbp (approved):

PR #449, opened 2023-02-15
feat(espisodes_list): 12432 future episodes on timeline should be displayed distinctly

Akiyamka:
Currently back-end not contain `forecasted` flag, so for local test you need go to 
https://github.com/konturio/disaster-ninja-fe/blob/d4a35047b4fac10f49f88d25d2c3eb46abdde38c/src/features/event_episodes/atoms/episodesResource.ts  line 43, and replace fallback to `true`


propakov (approved):
👍

src/features/event_episodes/atoms/episodesResource.ts
```diff
@@ -34,10 +34,14 @@ export const episodesResource = createAsyncAtom(
...
+      // Adapter:
+      // - add missing ids
+      // - set default value for forecasted flag;
       return responseData.map((e) => {
-        if (!e.id) e.id = 'temp_' + nanoid(6);
-        return e;
+        const cp = { ...e };
+        if (!cp.id) cp.id = 'temp_' + nanoid(6);
+        if (!cp.forecasted) cp.forecasted = false;
+        return cp;
```
IlyaIzr: i mean it's okay but takes time to guess what's `cp`, and i still couldn't guess it but i see what it does

src/features/event_episodes/atoms/episodesResource.ts
```diff
@@ -34,10 +34,14 @@ export const episodesResource = createAsyncAtom(
...
 
-      // Episodes not have any ids
+      // Adapter:
+      // - add missing ids
+      // - set default value for forecasted flag;
       return responseData.map((e) => {
-        if (!e.id) e.id = 'temp_' + nanoid(6);
-        return e;
+        const cp = { ...e };
+        if (!cp.id) cp.id = 'temp_' + nanoid(6);
```
IlyaIzr: just to be sure - id couldn't be `0`? 😆

src/features/event_episodes/atoms/episodesResource.ts
Akiyamka: `cp`[ is common shortcut for unix-like users ](https://linuxize.com/post/how-to-copy-files-and-directories-in-linux/), but if you not familiar with it, I will replace it to `copy`

src/features/event_episodes/atoms/episodesResource.ts
Akiyamka: nice catch, fixed

src/features/event_episodes/components/EpisodesTimeline/EpisodesTimeline.tsx
```diff
@@ -43,19 +65,16 @@ export function EpisodesTimeline({ episodes }: { episodes: Episode[] }) {
...
-      }
-    },
-    [],
-  );
+  const onSelect = useCallback((selection) => {
+    if (selection.length > 1) {
+      return; // its cluster
+    }
+    if (selection[0] !== undefined) {
+      eventEpisodesController.setCurrentEpisode(String(selection[0].id));
```
dqunbp:
```suggestion
    const [item] = selection
    if (item !== undefined) {
      eventEpisodesController.setCurrentEpisode(String(item.id));
```

dqunbp (approved):

PR #452, opened 2023-02-16
docs(core) - docs for tooltip added

Akiyamka:


src/core/tooltips/rfc.md
```diff
@@ -0,0 +1,87 @@
...
+    selectedFeatures.map((f) => f.properties.countryName).join(' ,'),
+  );
+});
+```
+
+### How service works
+
+```ts
+interface TooltipService {
+  init(mountPoint: Element) => void;
```
Akiyamka:
```suggestion
```
library use body for automatically create container
https://github.dev/floating-ui/floating-ui

src/core/tooltips/rfc.md
```diff
@@ -0,0 +1,87 @@
+## Tooltip Service
+
+This service allow us show tooltips in different places of our application.
```
IlyaIzr:
```suggestion
This service allows us to show tooltips in different places of our application.
```

src/core/tooltips/rfc.md
```diff
@@ -0,0 +1,87 @@
...
+Underhood it uses `@konturio/ui-kit` tooltip component for show tooltip in theme style
```
IlyaIzr: "to show" or "for showing"

Also i think all ui-kit components are style-compatible, no?

src/core/tooltips/rfc.md
```diff
@@ -0,0 +1,87 @@
...
+
+Tooltip:
+
+- shows by hover-in and hides by hover-out events (after delay)
+- Not interactive
+- Can contain only text
```
IlyaIzr: what about rich text\markdown?

src/core/tooltips/rfc.md
```diff
@@ -0,0 +1,87 @@
...
+In that case you need import special components and use them
+
+```jsx
+import { Tooltip, TooltipTrigger, TooltipContent } from '~core/tooltip';
+
+function Sidebar() {
+  return (
+    <nav>
+      <Tooltip>
+        {/* Where show */}
```
IlyaIzr:
```suggestion
        {/* Where to show */}
```

src/core/tooltips/rfc.md
```diff
@@ -0,0 +1,87 @@
...
+        <TooltipTrigger>
+          <button>Some button with tip</button>
+        </TooltipTrigger>
+        {/* What show */}
```
IlyaIzr:
```suggestion
        {/* What to show */}
```

src/core/tooltips/rfc.md
```diff
@@ -0,0 +1,87 @@
...
+
+In that case you need import special components and use them
+
+```jsx
+import { Tooltip, TooltipTrigger, TooltipContent } from '~core/tooltip';
+
+function Sidebar() {
+  return (
+    <nav>
+      <Tooltip>
```
IlyaIzr, line 26: Looks like TooltipContext or TooltipWrap or smth

IlyaIzr: Not subject of this PR i guess

src/core/tooltips/rfc.md
```diff
@@ -0,0 +1,87 @@
...
+});
+```
+
+### How service works
+
+```ts
+interface TooltipService {
+  init(mountPoint: Element) => void;
+  showTooltip: ({ x: number, y: number }, content: string) => number);
+  updateTooltip: ({ x: number, y: number }, content: string) => number);
```
IlyaIzr:
```suggestion
type TooltipId = number;

interface TooltipService {
  init(mountPoint: Element) => void;
  showTooltip: ({ x: number, y: number }, content: string) => TooltipId);
  updateTooltip: ({ x: number, y: number }, content: string) => TooltipId);
```

src/core/tooltips/rfc.md
```diff
@@ -0,0 +1,87 @@
+## Tooltip Service
+
+This service allow us show tooltips in different places of our application.
+Underhood it uses `@konturio/ui-kit` tooltip component for show tooltip in theme style
+
+Tooltip:
+
+- shows by hover-in and hides by hover-out events (after delay)
```
propakov:
```suggestion
- Shows by hover-in and hides by hover-out events (after delay)
```

src/core/tooltips/rfc.md
```diff
@@ -0,0 +1,87 @@
...
+- Not interactive
+- Can contain only text
+
+## How to use
+
+We assume that all cases of using a tooltip can be reduced to two cases
+
+### We have an element around which a tooltip should be shown
+
+In that case you need import special components and use them
```
propakov:
```suggestion
In that case you need to import special components and use them
```

src/core/tooltips/rfc.md
```diff
@@ -0,0 +1,87 @@
...
+  if (currentTooltipId !== null) {
+    tooltips.closeTooltip(currentTooltipId);
+    currentTooltipId = null;
+  }
+
+  // Show new tooltip
+  currentTooltipId = tooltips.showTooltip(
+    // Where show
+    { x: event.point.x, y: event.point.y },
+    // What show
```
propakov:
```suggestion
    // Where to show
    { x: event.point.x, y: event.point.y },
    // What to show
```

src/core/tooltips/rfc.md
dqunbp: Less words === Simpler

src/core/tooltips/rfc.md
Akiyamka: > Also i think all ui-kit components are style-compatible, no?

Sure. this text sounds like it doesn't?

src/core/tooltips/rfc.md
```diff
@@ -0,0 +1,86 @@
...
+  });
+
+  // Close previous tooltip
+  if (currentTooltipId !== null) {
+    tooltips.closeTooltip(currentTooltipId);
+    currentTooltipId = null;
+  }
+
+  // Show new tooltip
+  currentTooltipId = tooltips.showTooltip(
```
propakov: so we won't use mapbox tooltips on map?

src/core/tooltips/rfc.md
```diff
@@ -0,0 +1,86 @@
...
+    selectedFeatures.map((f) => f.properties.countryName).join(' ,'),
+  );
+});
+```
+
+### How service works
+
+```ts
+interface TooltipService {
+  showTooltip: ({ x: number, y: number }, content: string) => number);
```
dqunbp: In some cases it will be useful to pass `element` to `showTooltip`, for example, VisTimeline has a hover callback that provides an `element` arg

src/core/tooltips/rfc.md
```diff
@@ -0,0 +1,86 @@
...
+
+In that case you need to import special components and use them
+
+```jsx
+import { Tooltip, TooltipTrigger, TooltipContent } from '~core/tooltip';
+
+function Sidebar() {
+  return (
+    <nav>
+      <Tooltip>
```
dqunbp, line 26: If we need text-only content, we can use a simpler design

```tsx
<Tooltip content="string-only-content">
  {/* Where to show */}
  <button>Some button with tip</button>
</Tooltip>
```

The current design is more compatible with `Popup` which can contain any content

dqunbp, line 26: According to the design, the tooltip can be sized, regular and small, we need some prop for the sizing
image

dqunbp, line 26: The tooltip open status will be uncontrolled by default, it will be useful to have an `open` prop for controlled usage.

src/core/tooltips/rfc.md
dqunbp: I think yes, all app tooltips will work the same way

src/core/tooltips/rfc.md
IlyaIzr: yep

src/core/tooltips/rfc.md
IlyaIzr: Less words === more interpretations 😄

src/core/tooltips/rfc.md
Akiyamka: @propakov 
I think it would be easier to maintain one single tooltip service. Is there a reason why mapbox tooltips are better?

Komzpa: Are you talking about tooltips or map popups really? Map popups have different behavior from tooltips.

src/core/tooltips/rfc.md
Akiyamka: It's better to keep like this.
I can imagine case when you want pass <Trans>  component, or <Markdown> to tootip content.
Also in feature we probably add settings for content and trigger, and it will be cool that we keep it separately

src/core/tooltips/rfc.md
Akiyamka: Map have "popups".
It can be used for showing tooltips.
https://docs.mapbox.com/mapbox-gl-js/example/popup-on-hover/

Or classical popups 
https://docs.mapbox.com/mapbox-gl-js/example/popup-on-click/

src/core/tooltips/rfc.md
Akiyamka: I'am not see any way how it can be intepretated here in wrong way

src/core/tooltips/rfc.md
Akiyamka: For what case you need that?
In case you need to open/close it manually you want to use service directly as in example with map

src/core/tooltips/rfc.md
propakov: they just sticks to map position, and draw on canvas, so that's the difference

propakov: agree that maintaining single service is easier if you are ok to go without sticking to the map

propakov (approved):

src/core/tooltips/rfc.md
dqunbp: For example, show a tooltip on some component when hovering some another

Akiyamka: Sounds like broken ux : )
But if you really need that you still can do it:
```ts
triggerElement.addEventListener('pointer', (e) => {
  tooltip.show(
      ElementWhereToShow,
      'Some text'
    );
})
  ```

dqunbp (approved):

PR #453, opened 2023-02-17
fix(layers_in_area): 14718, 14900 - smart city layers display and focus geometry style wasn't shown properly

IlyaIzr:
changes:
1. warning added for case when `layersToRetrieveWithoutGeometryFilter` added to `/details` request without attached geometry. Happens on Smart City app because `focused-geometry` layer was added to list of global layers + enabled, yet no focused geometry exists by default for Smart City app
2. added another type for `LayerInAreaDetails` when minimal amount of data is provided and handled
3. added skips and warnings for case when incoming layer `legend` or `source` are missing and trying to overwrite existing layers

result:
console error on smart city app default opening fixed
request to remove `focused-geometry` from global layers was done
we found out that the minimal possible amount of data for layer response and accounted it

### DIFF
--- a/src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
+++ b/src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
@@ -11,7 +11,6 @@ import {
 } from './areaLayersDetailsResourceAtomCache';
 import type { DetailsRequestParams } from './types';
 
-// Why we add layersToRetrieveWithGeometryFilter if no geometry provided?
 export const areaLayersDetailsParamsAtom = createAtom(
   {
     enabledLayersAtom,
--- a/src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
+++ b/src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
@@ -11,6 +11,7 @@ import {
 } from './areaLayersDetailsResourceAtomCache';
 import type { DetailsRequestParams } from './types';
 
+// Why we add layersToRetrieveWithGeometryFilter if no geometry provided?
 export const areaLayersDetailsParamsAtom = createAtom(
   {
     enabledLayersAtom,
@@ -102,6 +103,11 @@ export const areaLayersDetailsParamsAtom = createAtom(
 
     if (focusedGeometry) {
       newState.geoJSON = focusedGeometry.geometry;
+    } else if (newState.layersToRetrieveWithoutGeometryFilter?.length) {
+      console.warn(
+        'missing geometry for layers ',
+        newState.layersToRetrieveWithoutGeometryFilter,
+      );
     }
 
     return newState;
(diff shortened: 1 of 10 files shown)


src/features/layers_in_area/atoms/areaLayersLegendsAndSources.ts
```diff
@@ -11,9 +11,12 @@ import type { LayerInAreaDetails } from '../types';
 import type { LayerSource } from '~core/logical_layers/types/source';
 import type { LayerLegend } from '~core/logical_layers/types/legends';
 
-function convertDetailsToSource(response: LayerInAreaDetails): LayerSource {
-  /* Typescript makes me sad sometimes T.T */
-  if ('url' in response.source) {
+function convertDetailsToSource(response: LayerInAreaDetails): LayerSource | null {
+  if (!('source' in response)) {
```
dqunbp: Can we use typeguards?

Something like this:

```ts
type SourceContainer = GeoJSONSourceSourceContainer | TileSourceContainer


const isSourceContainer = (data: LayerInAreaDetails): data is SourceContainer => 'source' in data

const isTileSourceContainer = (data: LayerInAreaDetails): data is TileSourceContainer => {
  const casted = data as TileSourceContainer
  return casted?.source?.type === 'vector' || casted?.source?.type === 'raster'
}

function convertDetailsToSource(response: LayerInAreaDetails): LayerSource | null {
  if (!isSourceContainer(response)) {
    return null;
  }

  if (isTileSourceContainer(response)) {
    ...
```

src/features/layers_in_area/utils/legendFormatter.ts
```diff
@@ -21,10 +21,8 @@ function convertBivariateColorsToSteps(
     });
 }
 
-export function legendFormatter(
-  details: LayerInAreaDetails,
-): LayerLegend | null {
-  if (!details.legend) {
+export function legendFormatter(details: LayerInAreaDetails): LayerLegend | null {
+  if (!('legend' in details) || !details.legend) {
```
dqunbp: Same with typeguards here, e.g.:

```ts
if(!isSourceContainer(details)) {
    ...

if(isBivariateLegend(details)) {
    ...
```

src/features/layers_in_area/atoms/areaLayersLegendsAndSources.ts
IlyaIzr: for readability? I guess we can

IlyaIzr: I believe casting type intentionally when it can be avoided is bad.

Avoiding that using slightly different approach doesn't work at our typescript at least.
Therefore i wouldn't do that

image

src/features/layers_in_area/atoms/areaLayersDetailsResource/areaLayersDetailsParamsAtom.ts
```diff
@@ -102,6 +102,11 @@ export const areaLayersDetailsParamsAtom = createAtom(
 
     if (focusedGeometry) {
       newState.geoJSON = focusedGeometry.geometry;
+    } else if (newState.layersToRetrieveWithoutGeometryFilter?.length) {
+      console.warn(
+        'missing geometry for layers ',
```
Akiyamka:
```suggestion
        'Layers require geometry, but geometry missing in payload',
```

src/features/layers_in_area/types.ts
```diff
@@ -38,6 +38,12 @@ export interface TileSourceContainer {
...
+export interface MinimalResponse {
+  id: string;
+  ownedByUser: boolean;
+}
+
 export type LayerInAreaDetails =
   | GeoJSONSourceSourceContainer
-  | TileSourceContainer;
+  | TileSourceContainer
+  | MinimalResponse;
```
Akiyamka: What is MinimalResponse? In what case we can get this? What kind of layers?

Akiyamka (changes requested):
give better name for "MinimalResponse"

Akiyamka (approved):

PR #459, opened 2023-02-21
doc(focused_geometry_editor) - Create docs for focused_geometry_editor feature

IlyaIzr:


src/features/focused_geometry_editor/readme.md
```diff
@@ -0,0 +1,30 @@
...
+```ts
+import('~features/focused_geometry_editor/').then(({ initFocusedGeometry }) =>
+  initFocusedGeometry(),
+);
+```
+
+#### On enabling
+
+`isEditorActiveAtom` sets to `true`, this triggers `focusedGeometryEditorAtom`. It hides focused geometry layer, if any geometry existed. It also starts listening to any uploaded files, so that if user uploads geometry - he would be able to edit it right away.
+`toolboxAtom` being enabled for all editor modes and also gets a callback action for disbling actions.
```
propakov:
```suggestion
`toolboxAtom` being enabled for all editor modes and also gets a callback action for disabling actions.
```

propakov (approved):

PR #465, opened 2023-02-22
chore(locate_me): 15002 - Do not switch off data collection if Locate me is tapped

IlyaIzr:
Removing `locate_me` feature from excluding toolbar groups (meaning activating it would not auto disable any other tool)


Akiyamka (approved):
LGFM

PR #467, opened 2023-02-22
doc(map_ruler): 13741 - doc for map_ruler

IlyaIzr:


src/features/map_ruler/readme.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+### How it works
+
+Root index file creates layers renderer instance and creates logical layer atom via `createLogicalLayerAtom` (it also adds layer to the layers registry).
+Then toolbars control being described, and the control logic is simple - enable logical layer on activation, disable on deactivation
+
+#### Map ruler renderer
+
+This features renderer is a custom renderer ([read more about renders](https://github.com/konturio/disaster-ninja-fe/blob/main/src/core/logical_layers/rfc.md#layers-renderers)).
```
Akiyamka:
```suggestion
This features renderer is a custom renderer ([read more about renderers](https://github.com/konturio/disaster-ninja-fe/blob/main/src/core/logical_layers/rfc.md#layers-renderers)).
```

src/features/map_ruler/readme.md
```diff
@@ -0,0 +1,28 @@
...
+On mount:
+
+1. DeckGl config is used to create new map layer. The point of this is to describe what to draw, which way and how to handle user interaction events. You can read more about (Deck Gl geojson layer props)[https://deck.gl/docs/api-reference/layers/geojson-layer], (Composite layer props)[https://deck.gl/docs/api-reference/core/composite-layer] to understand configs described in `MapRulerRenderer`. You can also read about [NebulaGl geojson layer](https://nebula.gl/docs/api-reference/layers/editable-geojson-layer) we're alterating and see it's [source code](https://github.com/uber/nebula.gl/blob/master/modules/layers/src/layers/editable-geojson-layer.ts) to see what methods we're changing
```
Akiyamka:
```suggestion
1. DeckGl config is used to create new map layer. The point of this is to describe what to draw, which way and how to handle user interaction events. You can read more about [Deck Gl geojson layer props](https://deck.gl/docs/api-reference/layers/geojson-layer), [Composite layer props](https://deck.gl/docs/api-reference/core/composite-layer) to understand configs described in `MapRulerRenderer`. You can also read about [NebulaGl geojson layer](https://nebula.gl/docs/api-reference/layers/editable-geojson-layer) we're alterating and see it's [source code](https://github.com/uber/nebula.gl/blob/master/modules/layers/src/layers/editable-geojson-layer.ts) to see what methods we're changing
```

Akiyamka (approved):
It is clearly written. Only typos need to be corrected.It is clearly written. Only typos need to be corrected.

PR #468, opened 2023-02-22
docs(layers_in_area): 13737 - readme added

Akiyamka:


src/features/layers_in_area/readme.md
```diff
@@ -0,0 +1,37 @@
+## Layer in area
+
+This feature manage all layers that intersecting area of interest (a.k.a focused geometry).
```
vkozio, line 3: This feature manages what layers are available in area of interest (a.k.a focused geometry).

^ Is it better explanation?

src/features/layers_in_area/readme.md
```diff
@@ -0,0 +1,37 @@
...
+Every time when geometry changes - this feature:
```
vkozio, line 4: probably needs more clarification what affects Layers in area

vkozio (commented):
see comments above

src/features/layers_in_area/readme.md
IlyaIzr: Yeah i think 'intersecting' is very specific word where's this feature manages all related to the AOI layers

src/features/layers_in_area/readme.md
IlyaIzr:
```suggestion
Every time when focused geometry changes - this feature:
```

src/features/layers_in_area/readme.md
```diff
@@ -0,0 +1,37 @@
...
+
+- generates and sends requests to the api
+- saves responses to a special cache
+- creates new logical layers
+- deletes the irrelevant layers
```
IlyaIzr: maybe smth like this?
```suggestion
- creates new logical layers if needed
- deletes the irrelevant layers if any
```

src/features/layers_in_area/readme.md
```diff
@@ -0,0 +1,37 @@
...
+
+### How it works
+
+`layersGlobalResource` - load all world wide available layers.
+`layersInAreaAndEventLayerResource` - load event related and focused geometry related layers
+
+They load only info about what layers available, but not layers itself.
+
+When info about available layers loaded - `areaLayersControlsAtom` creates logical layers for each,
+so after that user/other features can enable or disable those layers.
```
IlyaIzr:
```suggestion
so after that user or other features can enable or disable those layers.
```

src/features/layers_in_area/readme.md
```diff
@@ -0,0 +1,37 @@
...
+
+`areaLayersDetailsResource` - load data for **enabled** layers from above resources,
+that required for show layers legend, draw them on map, and so on.
```
IlyaIzr:
```suggestion
that required to show layers legend, draw them on map, and so on.
```
Also, isn't it all we get from details? Legend and source to generate legend and show layer on the mao?

src/features/layers_in_area/readme.md
Akiyamka: good point, i will use vkozio variant

src/features/layers_in_area/readme.md
Akiyamka: good point, i will use vkozio variant

PR #478, opened 2023-02-27
fix(core): 15075 DN FE fix layer details processing and TS types to match endpoint schema

vkozio:
https://kontur.fibery.io/Tasks/Task/DN-FE-fix-layer-details-processing-and-TS-types-to-match-endpoint-schema-15075


src/core/logical_layers/types/source.ts
```diff
@@ -23,3 +25,33 @@ export interface LayerTileSource {
 }
 
 export type LayerSource = LayerGeoJSONSource | LayerTileSource;
+
+export interface GeoJSONSourceSourceContainer {
```
Akiyamka: SourceSource

src/core/logical_layers/types/source.ts
```diff
@@ -23,3 +25,33 @@ export interface LayerTileSource {
...
+  id: string;
+  maxZoom: number;
+  minZoom: number;
+  source: TileSource;
+  legend?: LayerDetailsLegend;
+  ownedByUser: boolean;
+}
+export type LayerDetailsDTO = GeoJSONSourceSourceContainer | TileSourceContainer;
+
+export interface LayerSummaryDto {
```
Akiyamka, line 42: We have so many entities around layers, maybe we shouldn't create a new one - `LayerSummary`.
it has not been used anywhere before and it is not clear what it refers to

Akiyamka (approved):

PR #479, opened 2023-02-27
docs(layers_panel): 13738 layers_panel feature

vkozio:


src/features/layers_panel/readme.md
```diff
@@ -0,0 +1,27 @@
+## Layers Panel
+
+This feature adds layers panel UI with layers tree.
```
Akiyamka: ...layers tree that generated from layers that was added to layers registry

Akiyamka (approved):

PR #490, opened 2023-03-06
docs(events_list): add docs for events_list

dqunbp:
https://kontur.fibery.io/Tasks/Task/Create-docs-for-events_list-feature-(712-lines)-13731

### DIFF
--- a/src/features/events_list/readme.md
+++ b/src/features/events_list/readme.md
@@ -73,7 +73,7 @@ The `Analytics` component displays statistics for an event, such as the affected
 
 The `autoSelectEvent` atom is responsible for automatically selecting an event when the `events_list` feature is loaded.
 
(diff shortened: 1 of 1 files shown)


src/features/events_list/readme.md
```diff
@@ -0,0 +1,101 @@
...
+- `openFullState`: A function that is called when the user clicks the call-to-action button.
+- `currentEventId`: The ID of the currently selected event.
+
+# Atoms
+
+## `autoSelectEvent`
+
+The `autoSelectEvent` atom is responsible for automatically selecting an event when the `events_list` feature is loaded.
+
+It accomplishes this by subscribing to the `eventListResourceAtom` and `currentEventResourceAtom` atoms. When `eventListResourceAtom` emits a new value, `autoSelectEvent` checks if the `currentEventResourceAtom` has a null value. If `currentEventResourceAtom` is null, `autoSelectEvent` selects the first event in the `eventListResourceAtom`.
```
Akiyamka: Important detail - this happens only in case if scheduledAutoSelect was switched to true state before.
This needed because we have cases when event not selected and this is ok (sse when scheduledAutoSelect switched)

Akiyamka (approved):

PR #494, opened 2023-03-08
docs(event_episodes): add feature docs

dqunbp:
https://kontur.fibery.io/Tasks/Task/Create-docs-for-event_episodes-feature-(296-lines)-13730

### DIFF
--- a/src/features/event_episodes/readme.md
+++ b/src/features/event_episodes/readme.md
@@ -18,7 +18,10 @@ A component that displays the timeline of episodes. It receives a list of episod
 
 ### episodesResource
 
-An atom that stores a list of episodes associated with the selected disaster. The atom loads data from the server and provides an interface for getting this data. `episodesResource` depends on `currentEventAtom` trough the `episodesResourceDependencyAtom`.
+An atom that stores a list of episodes associated with the selected disaster. The atom loads data from the server and provides an interface for getting this data. `episodesResource` depends on `currentEventAtom` through the `episodesResourceDependencyAtom`.
+
+`currentEventAtom` stores the ID of the current (selected) event.
+
 When the `episodesResourceDependencyAtom` changes, the `episodesResource` atom is invalidated and the data is reloaded from the server.
 
 ### episodesTimeline
@@ -42,4 +45,4 @@ The `autoClearCurrentEpisode` function subscribes to the `episodesResource` atom
 
 ### eventEpisodesController
 
-A controller that provides methods for managing the selected episode, EpisodesTimelinePanel, and timeline. The controller links methods provided by various atoms and provides them as a single interface.
+A controller that provides methods for managing the selected episode, EpisodesTimelinePanel, and timeline. The controller links methods are provided by various atoms and are collected into a single interface.
--- a/src/features/event_episodes/readme.md
+++ b/src/features/event_episodes/readme.md
@@ -1,0 +1,45 @@
+# Event Episodes
+
+## Introduction:
+
+This feature allows the user to view the episodes of a disaster on a timeline, select a specific episode or cluster of episodes, and view data related to the selected episode.
+
+## Components
+
+### EpisodesTimelinePanel
+
+A panel containing the timeline of episodes. The user can open and close the panel. If the app is running on a mobile device, the panel opens in a modal window.
+
+### EpisodesTimeline
+
+A component that displays the timeline of episodes. It receives a list of episodes as `episodes` prop and displays them on the timeline. The user can select a specific episode or cluster of episodes to view it on the map. The component also supports data-driven styles for episodes and clusters of episodes.
+
+## Atoms
+
+### episodesResource
+
+An atom that stores a list of episodes associated with the selected disaster. The atom loads data from the server and provides an interface for getting this data. `episodesResource` depends on `currentEventAtom` trough the `episodesResourceDependencyAtom`.
+When the `episodesResourceDependencyAtom` changes, the `episodesResource` atom is invalidated and the data is reloaded from the server.
+
+### episodesTimeline
+
+An atom that stores timeline settings, such as stacking and clustering. The atom provides an interface for changing these settings and for controlling the imperative timeline API.
+
+### episodeToFocusedGeometry
+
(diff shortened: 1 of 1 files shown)


src/features/event_episodes/readme.md
```diff
@@ -0,0 +1,45 @@
...
+
+### EpisodesTimeline
+
+A component that displays the timeline of episodes. It receives a list of episodes as `episodes` prop and displays them on the timeline. The user can select a specific episode or cluster of episodes to view it on the map. The component also supports data-driven styles for episodes and clusters of episodes.
+
+## Atoms
+
+### episodesResource
+
+An atom that stores a list of episodes associated with the selected disaster. The atom loads data from the server and provides an interface for getting this data. `episodesResource` depends on `currentEventAtom` trough the `episodesResourceDependencyAtom`.
```
amurKontur: change _trough_ to _**through**_

src/features/event_episodes/readme.md
```diff
@@ -0,0 +1,45 @@
...
+
+### episodesTimeline
+
+An atom that stores timeline settings, such as stacking and clustering. The atom provides an interface for changing these settings and for controlling the imperative timeline API.
+
+### episodeToFocusedGeometry
+
+The `episodeToFocusedGeometry` atom is used to convert selected episode into a focused geometry.
+When a user selects a new episode, the `currentEpisodeAtom` is updated with the selected episode. Then `currentEpisodeAtom` retrieves the geometry for the selected episode.
+Then it merges the episode geometry feature collection to single geometry, using the `mergeFeatureCollection` function, and then sets the focused geometry using the `setFocusedGeometry` function from the `focusedGeometryAtom`. Finally, the function returns the `currentEpisodeData` object.
```
amurKontur, line 35: what is focusedGeometryAtom? A feature, a class?
What is currentEpisodeData object about? Maybe more details to add, as it was done above

src/features/event_episodes/readme.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Introduction:
+
+This feature allows the user to view the episodes of a disaster on a timeline, select a specific episode or cluster of episodes, and view data related to the selected episode.
+
+## Components
+
+### EpisodesTimelinePanel
+
+A panel containing the timeline of episodes. The user can open and close the panel. If the app is running on a mobile device, the panel opens in a modal window.
```
amurKontur, line 11: from which resolution does it think that it is a mobile device? more details on a mobile device defining

src/features/event_episodes/readme.md
amurKontur: more details on `currentEventAtom` and `episodesResourceDependencyAtom`. What is inherited from what and what just depends on each other

src/features/event_episodes/readme.md
```diff
@@ -0,0 +1,45 @@
...
+
+### episodeToFocusedGeometry
+
+The `episodeToFocusedGeometry` atom is used to convert selected episode into a focused geometry.
+When a user selects a new episode, the `currentEpisodeAtom` is updated with the selected episode. Then `currentEpisodeAtom` retrieves the geometry for the selected episode.
+Then it merges the episode geometry feature collection to single geometry, using the `mergeFeatureCollection` function, and then sets the focused geometry using the `setFocusedGeometry` function from the `focusedGeometryAtom`. Finally, the function returns the `currentEpisodeData` object.
+
+### autoClearCurrentEpisode
+
+`autoClearCurrentEpisode` clears the currently selected episode when a new disaster is selected. This helps to prevent confusion when the user switches between different disasters.
```
amurKontur, line 36: does it set = null? Maybe some info on how exactly clearing happens

src/features/event_episodes/readme.md
```diff
@@ -0,0 +1,45 @@
...
+
+**Behavior**
+The `autoClearCurrentEpisode` function subscribes to the `episodesResource` atom. When the `episodesResource` changes, the function checks if the current episode is associated with the selected disaster. If it is not, the function updates the `currentEpisodeAtom` with a null value, effectively clearing the current episode selection.
```
amurKontur, line 42: what do you mean by subscribes? wouldn't it be better if we used something like follows/tracks?

src/features/event_episodes/readme.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Controller
+
+### eventEpisodesController
+
+A controller that provides methods for managing the selected episode, EpisodesTimelinePanel, and timeline. The controller links methods provided by various atoms and provides them as a single interface.
```
amurKontur: Change _The controller links methods provided by various atoms and provides them as a single interface._ to 
_**The controller links methods are provided by various atoms and are collected into a single interface.**_

src/features/event_episodes/readme.md
dqunbp: It's easy to see from the code.
It is more important to know that there is such a behavior than to point to any specific values.

src/features/event_episodes/readme.md
dqunbp: It is enough to know about these atoms how they depend on each other and how the `episodesResource` atom depends on them.
`currentEventAtom` is part of the shared state, its documentation should be in `src/core/shared_state/`

To understand why these dependencies are created, it is necessary to understand how all `resource` atoms work in the application, which should be described at a higher level of documentation.

src/features/event_episodes/readme.md
dqunbp: Documentation describing `focusedGeometryAtom` should be in `src/core/focused_geometry/`, similar for `currentEpisodeAtom`.

To understand how `focusedGeometryAtom` works, you need to read the reatom documentation.

If we start to describe in detail all the entities involved in this way, we will describe half of the project here.
The documentation of this feature should contain its description directly.

src/features/event_episodes/readme.md
dqunbp: This is outside the scope of the documentation for this feature, to understand this you need to study the documentation of the reatom library.

Akiyamka (approved):

PR #497, opened 2023-03-10
fix(side_bar): use element based tooltips position

dqunbp:
Use element-based positioning for sidebar tooltips 
Extract NavButton component
Extract ToggleButton component

https://kontur.fibery.io/Tasks/Task/Wrong-tooltips-positioning-on-DN-sidebar-buttons-14909

### DIFF
--- a/src/features/side_bar/components/SideBar/NavButton.tsx
+++ b/src/features/side_bar/components/SideBar/NavButton.tsx
@@ -10,45 +10,39 @@ import type { AppRoute, CurrentRouteAtom } from '~core/router';
 
 type NavButtonProps = {
   route: AppRoute;
+  isVisible: boolean;
   minified: boolean;
   checkRouteVisibility: (route: AppRoute, currentRoute: AppRoute | null) => boolean;
   showTooltip?: boolean;
   currentRouteAtom: CurrentRouteAtom;
-  getAbsoluteRoute: (path: string) => string;
+  getAbsoluteRoute: (path: string | AppRoute) => string;
 };
 
 export function NavButton({
   route,
+  isVisible,
   minified,
-  checkRouteVisibility,
   showTooltip,
   currentRouteAtom,
-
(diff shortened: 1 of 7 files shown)


src/features/side_bar/components/SideBar/NavButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+  checkRouteVisibility,
+  showTooltip,
+  currentRouteAtom,
+
+  getAbsoluteRoute,
+}: NavButtonProps) {
+  const [searchString] = useAtom(searchStringAtom);
+  const [currentRoute] = useAtom(currentRouteAtom);
+
+  const isVisible = checkRouteVisibility(route, currentRoute);
```
Akiyamka: Why not pass isVisible as boolean prop? Right now this function invokes on every render

src/features/side_bar/components/SideBar/NavButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  const ref = useRef<HTMLDivElement>(null);
+  const [isTooltipOpen, setIsTooltipOpen] = useState(false);
+
+  if (!isVisible) return null;
+
+  const to = getAbsoluteRoute(
```
Akiyamka: Looks like router logic

src/features/side_bar/components/SideBar/NavButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+    route.parentRoute
+      ? `${route.parentRoute}/${route.slug}${searchString}`
+      : `${route.slug}${searchString}`,
+  );
+
+  const className = cn(
```
Akiyamka: any element  may have className prop. If possible - write this logic in jsx, in you need to move it in variable - give them some more unique name

Akiyamka (approved):

PR #498, opened 2023-03-10
chore(localization): 15197 add check for untranlated strings count into GitHub worflow

Akiyamka:


scripts/i18n/utils/parser.ts
```diff
@@ -0,0 +1,80 @@
+import fs from 'node:fs/promises';
```
Komzpa, line 1: is there a preexisting parser of .po in the codebase?

Komzpa (approved):

scripts/i18n/utils/parser.ts
Akiyamka: we have https://www.npmjs.com/package/gettext-parser in dependencies, i can use it instead of our parser

PR #499, opened 2023-03-13
feat(mcda): 15238 - dynamic min max in mcda style

Akiyamka:
if we have layer where normalization disabled
we must take show all between min and max range.
in opposite case - between 0 and 1

### DIFF
--- a/src/features/mcda/atoms/mcdaCalculation/mapStyle.ts
+++ b/src/features/mcda/atoms/mcdaCalculation/mapStyle.ts
@@ -37,7 +37,7 @@ export function linearNormalization(layers: MCDAConfig['layers']) {
 export function createMCDAStyle(config: MCDAConfig) {
   const { good = DEFAULT_GREEN, bad = DEFAULT_RED } = config.colors;
   const [absoluteMin = 0, absoluteMax = 1] = config.layers.reduce((acc, l) => {
-    // Show full range of values between min max in normalization not enabled
+    // Show full range of values between min max if normalization not enabled
     const range: [number, number] = l.normalization === 'no' ? l.range : [0, 1];
     if (acc.length === 0) return range;
     acc[0] = Math.min(acc[0], range[0]);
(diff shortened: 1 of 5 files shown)


src/features/mcda/atoms/mcdaCalculation/mapStyle.ts
```diff
@@ -36,6 +36,14 @@ export function linearNormalization(layers: MCDAConfig['layers']) {
 
 export function createMCDAStyle(config: MCDAConfig) {
   const { good = DEFAULT_GREEN, bad = DEFAULT_RED } = config.colors;
+  const [absoluteMin = 0, absoluteMax = 1] = config.layers.reduce((acc, l) => {
+    // Show full range of values between min max in normalization not enabled
```
dqunbp:
```suggestion
    // Show full range of values between min max if normalization not enabled
```

src/features/mcda/atoms/mcdaCalculation/mapStyle.ts
```diff
@@ -36,6 +36,14 @@ export function linearNormalization(layers: MCDAConfig['layers']) {
...
+    const range: [number, number] = l.normalization === 'no' ? l.range : [0, 1];
+    if (acc.length === 0) return range;
```
dqunbp, line 42: `acc. length` initially always is zero?

dqunbp (approved):

src/features/mcda/atoms/mcdaCalculation/mapStyle.ts
```diff
@@ -45,10 +53,14 @@ export function createMCDAStyle(config: MCDAConfig) {
       'fill-color': [
         'let',
         'mcdaResult',
-        ['to-number', linearNormalization(config.layers), -1], // falsy values become -1
+        ['to-number', linearNormalization(config.layers), -9999], // falsy values become -1
```
Akiyamka:
```suggestion
        ['to-number', linearNormalization(config.layers), -9999], // falsy values become -9999
```

PR #504, opened 2023-03-15
feat(core): 14199 utilize dn2 be boundaries endpoint to retrieve administrative boundaries

Akiyamka:


vkozio (approved):

src/components/ConnectedMap/ConnectedMap.tsx
```diff
@@ -121,7 +120,7 @@ export function ConnectedMap({ className }: { className?: string }) {
 
   return (
     <Map
-      accessToken={accessToken}
+      accessToken={''}
```
dqunbp, line 123: Do we need this prop anymore?

Akiyamka: Not needed, but this out of scope of this task

PR #505, opened 2023-03-17
fix(core): 14439 filter-out-layers-that-can't-be-shown

vkozio:
https://kontur.fibery.io/Tasks/Task/DN2-FE.-Utilize-layer-type-property-to-filter-out-layers-that-can't-be-shown-14439

filter unsupported layers (hardcoded list till BE return correct type)
apiClient fix
add missing atom names
get default basemap from api

### DIFF
--- a/src/core/api_client/apiClient.ts
+++ b/src/core/api_client/apiClient.ts
@@ -444,10 +444,12 @@ export class ApiClient {
         );
       }
 
-      if (!requestConfig.headers || !requestConfig.headers.Authorization) {
-        requestConfig.headers = {
-          Authorization: `Bearer ${this.token}`,
-        };
+      if (!requestConfig.headers) {
+        requestConfig.headers = {};
+      }
+
+      if (!requestConfig.headers.Authorization) {
+        requestConfig.headers.Authorization = `Bearer ${this.token}`;
       }
     }
 
--- a/src/core/app/init.ts
+++ b/src/core/app/init.ts
@@ -1,13 +1,18 @@
-import { updateAppConfig } from '~core/app_config';
+import { updateAppConfig, updateAppConfigOverrides } from '~core/app_config';
 import { apiClient } from '~core/apiClientInstance';
 import { urlEncoder, urlStoreAtom } from '~core/url_store';
 import { authClientInstance } from '~core/authClientInstance';
 import { i18n } from '~core/localization';
+import {
+  findBasemapInLayersList,
+  getBasemapFromDetails,
+} from '~core/logical_layers/basemap';
 import { onLogin } from './authHooks';
 import { defaultUserProfileData } from './user';
 import { runAtom } from './index';
+import type { LayerDetailsDto } from '~core/logical_layers/types/source';
 import type { UrlData } from '~core/url_store';
-import type { AppConfiguration } from '~core/app/types';
+import type { AppDto } from '~core/app/types';
 
 export async function appInit() {
   // keep initial url before overwriting by router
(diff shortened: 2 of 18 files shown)


src/core/api_client/apiClient.ts
```diff
@@ -444,10 +444,12 @@ export class ApiClient {
...
-      if (!requestConfig.headers || !requestConfig.headers.Authorization) {
-        requestConfig.headers = {
-          Authorization: `Bearer ${this.token}`,
-        };
+      if (!requestConfig.headers) {
+        requestConfig.headers = {};
+      }
+
+      if (!requestConfig.headers.Authorization) {
+        requestConfig.headers.Authorization = `Bearer ${this.token}`;
```
Akiyamka, line 452: Hard to understand what happens here.
Before that  if have pretty straight logic - if custom  header not set - use default headers  (with auth header)

Now it's unusual  
If headers not set - we don't set auth header, but if we have at least one - then set auth header.

To be more clear, we can name prop where we pass additional headers  as  `additionalHeaders` 
In node js you usually have and request.headers object where you can add (headers.add('...', '...')
what you want.

What you think about next approach: pass object for override or function for patch? (is widely used in react):

```ts
// Set defautls
const defaultHeaders = {
  Authorization: `Bearer ${this.token}`,
};

if (typeof requestConfig.headers === 'function') {
  // Patch it when needed
  requestConfig.headers = requestConfig.headers(defaultHeaders)
} else {
  // Fallback to defaults
  requestConfig.headers ??= defaultHeaders;
}
```

src/core/app/types.ts
```diff
@@ -76,24 +77,28 @@ export interface AppConfigGlobal {
 }
 
 // Unified config from /app/configuration
-export type AppConfiguration = {
+export interface AppDto {
```
Akiyamka, line 80:
```suggestion
export interface AppConfigDto {
```

src/core/app/init.ts
```diff
@@ -34,17 +39,32 @@ export async function appInit() {
...
-    initialState.layers = await getDefaultLayers(initialState.app);
+    initialState.layers = defaultLayers?.map((l) => l.id) ?? [];
   } else {
     // HACK: Remove KLA__ prefix from layers ids coming from url
     initialState.layers = initialState.layers.map((l) => l.replace(/^KLA__/, ''));
+
+    const basemapInUrl = findBasemapInLayersList(initialState.layers);
+    if (!basemapInUrl) {
+      initialState.layers.push(basemapLayerId);
+    }
```
Akiyamka, line 58: not clear what happens here:
Why we looking for all base map layers, but enable only one from them?

Akiyamka: This will enable basemap after page refresh in case user disable it. If it intended - better to describe this in comment. Also, looks like we forget to check case when default base map not listed app config (in case we want map without basemap in some app)
```suggestion
    /* Enable app default basemap layer if it not listed in url */
    const basemapInUrl = findBasemapInLayersList(initialState.layers);
    if (!basemapInUrl) {
      initialState.layers.push(basemapLayerId);
    }
```

src/core/app_config/index.ts
```diff
@@ -15,13 +15,19 @@ declare global {
 }
 
 const configs = {
-  global: {} as AppConfigGlobal,
-  custom: {} as AppConfiguration,
-  merged: {} as AppConfigEffective,
+  global: {} as AppConfigGlobal, // default from static json
+  custom: {} as AppDto, // updates from api
+  overrides: {} as Partial<AppConfigEffective>,
+  merged: {} as AppConfigEffective, // final merged
```
Akiyamka, line 21: Love this comments!

Akiyamka (approved):

src/core/api_client/apiClient.ts
vkozio: TBD in apiClient rewrite task

src/core/app/types.ts
vkozio: It's name of BE schema, we can review that in config refactor task

PR #510, opened 2023-03-29
feat(mcda): 15105-add-coloring-options-to-mcda-layer

Akiyamka:
Now user can describe any style using mapbox expressions

### DIFF
--- a/src/features/mcda/atoms/mcdaCalculation/mapStyle.ts
+++ b/src/features/mcda/atoms/mcdaCalculation/mapStyle.ts
@@ -118,11 +118,13 @@ export function createMCDAStyle(config: MCDAConfig) {
 
   const mcdaResult = linearNormalization(config.layers);
 
-  const layerStyle = {
+  const layerStyle: BivariateLayerStyle = {
     id: config.id,
     type: 'fill',
     layout: {},
     filter: filterSetup(config.layers),
+    // TODO - MCDA should have separate from bivariate renderer
+    // @ts-expect-error - any style props can be here
     paint: generateLayerPaint({
       colorsConfig: config.colors,
       mcdaResult,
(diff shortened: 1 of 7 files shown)


src/features/mcda/parser/adapters.ts
```diff
@@ -1,11 +1,14 @@
 import type { MCDAConfig } from '../types';
-import type { JsonMCDAv1, JsonMCDAv2, JsonMCDAv3 } from './types';
+import type { JsonMCDAv1, JsonMCDAv2, JsonMCDAv3, JsonMCDAv4 } from './types';
 
-export function firstVersionAdapter(json: JsonMCDAv1): MCDAConfig {
+export function firstVersionMigration(json: JsonMCDAv1): MCDAConfig {
```
vkozio, line 4: will naming like JsonMCDAv1Migration or MCDAv1Migration look more understandable?

src/features/mcda/atoms/mcdaCalculation/mapStyle.ts
```diff
@@ -34,8 +34,79 @@ export function linearNormalization(layers: MCDAConfig['layers']) {
...
+          ]
+        : expression;
+
+      return acc;
+    },
+    {} as Record<any, any>,
+  );
+}
+
+function generateLayerPaint(props: PaintProps) {
```
vkozio, line 99: No default case for switch - is it safe here?

vkozio (changes requested):

vkozio (approved):

src/features/mcda/parser/adapters.ts
Akiyamka: because these functions essentially describe the migration of older versions to newer ones

src/features/mcda/atoms/mcdaCalculation/mapStyle.ts
Akiyamka: Since the data is validated on the input against the types, it is safe. When we add new invariant in types - ts rise error here

PR #520, opened 2023-04-19
feat(live-sensors): 15603 frontend shouldnt duplicate live sensor data

Akiyamka:
- Added https mode for dev server - usefull for testing live_sensor feature because it works only in secured env
- Logic rewritten from atoms to pure classes, reactivity not needed here
- Tests added
- Fixed few additional bugs like - button still pressed if user disallow sensor usage, dublicate data sended
- Unique id added to every snapshot (every feature)
- Docs updated


src/features/live_sensor/SensorsSnapshotsSender.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  public start() {
+    this.sendFromQueue();
+  }
+
+  private running = true;
+  async sendFromQueue() {
+    // * Note: will not work in concurrent mode
+    while (this.running && this.sensorsRecords[0]) {
```
vkozio: Will it be more clear to check array length?

src/features/live_sensor/sensors/AppSensorAccelerometer.ts
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+
+  public stop() {
+    if (this.sensor) {
+      this.sensor.removeEventListener('reading', this.readingHandler);
+      this.sensor.removeEventListener('error', this.errorHandler);
+      this.sensor.stop();
+    }
+  }
```
vkozio: it's similar for several sensors - can be moved to parent class

vkozio (approved):
LGTM, but please check comments

src/features/live_sensor/sensors/AppSensorAccelerometer.ts
Akiyamka: unforutantelly parent class don't known about sensors from typescript perspective

src/features/live_sensor/SensorsSnapshotsSender.ts
Akiyamka: agree

PR #522, opened 2023-04-20
fix(core): 14901 update api client to more contemporary library

vkozio:


src/core/api_client/apiClient.ts
```diff
@@ -144,26 +144,24 @@ export class ApiClient {
...
   private async _tokenRefreshFlow() {
-    if (this.tokenExpirationDate) {
-      const diffTime = this.tokenExpirationDate.getTime() - new Date().getTime();
-      if (diffTime < 0) {
-        // token expired
-        this.resetAuth();
+    if (!this.tokenExpirationDate) {
+      return false;
+    }
+    const diffTime = this.tokenExpirationDate.getTime() - new Date().getTime();
```
Komzpa, line 154: I think this is going to fail if user's computer clock is wildly out of sync, you are comparing server time with client time here. No need to fix now other than maybe error message that clock is out of sync if the token is invalid right after receiving.

Akiyamka (approved):
Lgfm

src/core/api_client/apiClient.ts
vkozio: error message implemented, also we have additional step to refresh token and retry when request fails with 401

PR #528, opened 2023-05-04
fix(logical_layers): 14033 replace generating h3 id with reading it from feature property

Akiyamka:
Before we takes map zoom and hover coordinate for find out what hex index we looked at.
Since we have more complex logic for define h3 resolution it not work anymore, and because of this we just read h3 index from feature for draw it without guessing it.

Note that for correct work this task require changes on back-end side


src/core/logical_layers/renderers/BivariateRenderer.tsx
```diff
@@ -64,23 +64,15 @@ const convertFillColorToRGBA = (fillColor: FillColor, withTransparency = true):
...
-  resolution: number,
-): GeoJSON.Geometry => {
-  const h3 = geoToH3(
-    lngLat.lat,
-    lngLat.lng,
-    Math.min(resolution, MAX_BIVARIATE_GENERATED_RESOLUTION),
-  );
-  const h3Boundary = h3ToGeoBoundary(h3, true);
+function getH3Polygon(h3Index: string): GeoJSON.Geometry {
+  const h3Boundary = h3ToGeoBoundary(h3Index, true);
```
Komzpa: no, you're not supposed to generate it at all, just hover the original feature using maplibre's mechanisms.

src/core/logical_layers/renderers/BivariateRenderer/setTileScheme.ts
```diff
@@ -0,0 +1,9 @@
+import type { RasterSource, VectorSource } from 'maplibre-gl';
+
+/* https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#vector-scheme */
+export function setTileScheme(rawUrl: string, mapSource: VectorSource | RasterSource) {
```
vkozio: we already have it in /src/core/logical_layers/renderers/GenericRenderer.ts#L479
should we move it (and similar stuff from GenericRenderer) to utils?

src/core/logical_layers/renderers/BivariateRenderer/types.ts
```diff
@@ -0,0 +1,6 @@
+export type FillColor = {
```
vkozio: I'd move it to core/types and reuse in /src/utils/bivariate/bivariateColorThemeUtils.ts#L229
also it can be named more generic - ColorRGBA

vkozio (approved):
LGTM, but please check my comments

src/core/logical_layers/renderers/BivariateRenderer/types.ts
Akiyamka: agree, but looks like the for such common types we have better place in `src/types`

PR #529, opened 2023-05-10
fix(layers_panel): 15866 layer clear previously delegated legend after in unmounted

Akiyamka:


vkozio (approved):
LGTM (except english spelling and grammar)

PR #530, opened 2023-05-11
fix(common): 15889 - do not crash when session or local storage not available

Akiyamka:
Fix for application crashing when trying to use localStorage in incognito mode

### DIFF
--- a/src/utils/storage/index.ts
+++ b/src/utils/storage/index.ts
@@ -1,0 +1,79 @@
+/* In memory replacement for storage */
+class FallbackStorage implements Storage {
+  private storage = new Map();
+
+  setItem(key: string, value: string): void {
+    this.storage.set(key, value);
+  }
+
+  getItem(key: string): string | null {
+    return this.storage.get(key);
+  }
+
+  removeItem(key: string): void {
+    this.storage.delete(key);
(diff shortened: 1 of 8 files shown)


src/utils/storage/index.ts
```diff
@@ -0,0 +1,79 @@
...
+  get length() {
+    return this.storage.size;
+  }
+
+  key(index: number): string | null {
+    return this.storage.keys()[index] ?? null;
+  }
+}
+
+class StableStorage implements Storage {
```
vkozio, line 30: PersistentStorage will be more correct name

src/utils/storage/index.ts
```diff
@@ -0,0 +1,79 @@
+/* In memory replacement for storage */
+class FallbackStorage implements Storage {
```
vkozio, line 2: InMemoryStorage will be more descriptive

vkozio (approved):

src/utils/storage/index.ts
Akiyamka: Persistent - not that I mean here. This more like - fail free, errorless. It's not persistent in case of in fallback to inmemory implementation

PR #545, opened 2023-06-06
fix(core): 14286 token refresh fixes

vkozio:
fixes
https://kontur.fibery.io/Tasks/Task/Token-was-not-automatically-refreshed-after-it-was-expired-14286
https://kontur.fibery.io/Tasks/Task/User-is-being-logged-out-automatically-after-several-days-14450


src/core/apiClientInstance.ts
```diff
@@ -10,7 +10,7 @@ ApiClient.init({
   refreshTokenApiPath: `${appConfig.keycloakUrl}/auth/realms/${appConfig.keycloakRealm}/protocol/openid-connect/token`,
   keycloakClientId: appConfig.keycloakClientId,
   unauthorizedCallback(apiClient) {
-    apiClient.logout();
+    // TODO: implement for this case special login flow without reload
```
Akiyamka, line 13: I can already say that I was right? =)

Akiyamka (approved):

PR #584, opened 2023-10-25
feat(toolbar): update sidebar paddings

dqunbp:
simplify sidebar components and classes

https://kontur.fibery.io/Tasks/Task/Update-paddings-on-sidebar-16873


Akiyamka (changes requested):
1. Keep paddings around logo the same as before (must be the same as on prod right now)
2. Icons in sidebar jumping when I switching between collapsed and normal states

Akiyamka (approved):

PR #640, opened 2024-01-15
fix(auth): add logout if we have err with refresh token

dqunbp:
If we have any problems with refresh token, add logout Improve auth alert error message
Improve auth error messages parsing

https://kontur.fibery.io/Tasks/Task/invalid_grant-shown-to-users-in-the-loop-for-users-with-wrong-invalid-refresh-token-17578

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of unauthorized and forbidden errors, providing clearer messages to users when access is denied or sessions expire.
	- Enhanced error reporting for users, now displaying more informative messages when encountering specific request errors.

### DIFF
--- a/src/core/apiClientInstance.ts
+++ b/src/core/apiClientInstance.ts
@@ -9,7 +9,7 @@ export const apiClient = withSetupCheck(ApiClient, {
       switch (error.problem.kind) {
         case 'unauthorized':
         case 'forbidden':
-          alert('Session expired. Please login again');
+          alert('Access denied or your session expired. Please login again');
           goTo('/profile');
           location.reload();
           break;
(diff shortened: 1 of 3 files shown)


src/core/apiClientInstance.ts
```diff
@@ -9,7 +9,7 @@ export const apiClient = withSetupCheck(ApiClient, {
       switch (error.problem.kind) {
         case 'unauthorized':
         case 'forbidden':
-          alert('Session expired. Please login again');
+          alert('Access denied or your session expired. Please login again');
```
Komzpa, line 12: can we wrap this into i18n translation string?

Akiyamka (approved):

PR #655, opened 2024-02-14
feat(17481): Download button added in layers panel for MCDA layer

albaranau:
https://kontur.fibery.io/Tasks/Task/Add-Download-MCDA-flow-17481

https://www.figma.com/file/cFmWZmGioB52BDMoCofzpY/MCDA?type=design&node-id=12026-171898&mode=design&t=t7EABNwtf8umJwG9-0

- Download button added in layers panel for MCDA layer
image

## Summary by CodeRabbit

- **New Features**
	- Enhanced download functionality to support layers with the 'mcda' style type.
	- Improved logic to determine downloadable layers based on new conditions.

### DIFF
--- a/src/core/logical_layers/utils/logicalLayerFabric.ts
+++ b/src/core/logical_layers/utils/logicalLayerFabric.ts
@@ -125,7 +125,9 @@ export function createLogicalLayerAtom(
         isEnabled: get('enabledLayersAtom').has(id),
         isMounted: mounted.has(id),
         isVisible: !get('hiddenLayersAtom').has(id),
-        isDownloadable: asyncLayerSource.data?.source.type === 'geojson', // details.data.source.type === 'geojson'
+        isDownloadable:
+          asyncLayerSource.data?.source.type === 'geojson' || // details.data.source.type === 'geojson'
+          asyncLayerSource.data?.style?.type === 'mcda',
         settings: deepFreeze(asyncLayerSettings.data),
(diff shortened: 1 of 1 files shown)


src/core/logical_layers/utils/logicalLayerFabric.ts
```diff
@@ -200,8 +202,15 @@ export function createLogicalLayerAtom(
                 state.settings?.name || state.id || 'Disaster Ninja map layer'
```
vkozio: app name (Disaster Ninja) here and below should be taken from config
`configRepo.get().name`

vkozio (dismissed):

src/core/logical_layers/utils/logicalLayerFabric.ts
Komzpa: or just should not be

albaranau: There's the same issue with mcda download. 
Should it be changed for both? Maybe it really is better to keep it simple, since it's a fallback name?

> app name (Disaster Ninja) here and below should be taken from config `configRepo.get().name`

dqunbp (approved):

src/core/logical_layers/utils/logicalLayerFabric.ts
albaranau: Can someone review/resolve this conversation please? @Komzpa @dqunbp

Komzpa (approved):

PR #677, opened 2024-03-26
fix: 17740: select/deselect disaster after filters changes

IlyaLosikKontur:
https://kontur.fibery.io/Tasks/Task/Select-a-top-disaster-after-switching-feed-on-DN2-17741
https://kontur.fibery.io/Tasks/Task/Deselect-a-disaster-on-DN2-after-using-map-view-filter-17740

## Summary by CodeRabbit

- **Refactor**
	- Enhanced flexibility in event selection by allowing the clearing of the current event selection.
	- Improved event list auto-selection logic to prioritize top disaster events.

### DIFF
--- a/src/core/shared_state/currentEvent.ts
+++ b/src/core/shared_state/currentEvent.ts
@@ -10,7 +10,7 @@ export type CurrentEventAtomState = {
 
 export const currentEventAtom = createAtom(
   {
-    setCurrentEventId: (eventId: string) => eventId,
+    setCurrentEventId: (eventId: string | null) => eventId,
     focusedGeometryAtom,
   },
   ({ onAction, onChange }, state: CurrentEventAtomState = null) => {
--- a/src/features/events_list/atoms/autoSelectEvent.ts
+++ b/src/features/events_list/atoms/autoSelectEvent.ts
@@ -29,13 +29,17 @@ export const autoSelectEvent = createAtom(
         if (currentEvent?.id) {
           // This case happens when call for event by provided eventId didn't return event
           schedule((dispatch) =>
-            dispatch(
+            dispatch([
               currentNotificationAtom.showNotification(
                 'warning',
                 { title: i18n.t('event_list.no_event_in_feed') },
                 5,
               ),
-            ),
+              // select top disaster from the list, if it's not empty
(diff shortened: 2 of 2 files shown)


src/features/events_list/atoms/autoSelectEvent.ts
```diff
@@ -29,13 +29,17 @@ export const autoSelectEvent = createAtom(
...
+            dispatch([
               currentNotificationAtom.showNotification(
                 'warning',
                 { title: i18n.t('event_list.no_event_in_feed') },
                 5,
               ),
-            ),
+              // select top disaster from the list, if it's not empty
+              currentEventAtom.setCurrentEventId(
+                (eventListResource.data || [])[0]?.eventId || null,
```
dqunbp, line 40: Looks very unreadable, can we try to find the event first and then call `currentEventAtom.setCurrentEventId(`

src/core/shared_state/currentEvent.ts
```diff
@@ -10,7 +10,7 @@ export type CurrentEventAtomState = {
 
 export const currentEventAtom = createAtom(
   {
-    setCurrentEventId: (eventId: string) => eventId,
+    setCurrentEventId: (eventId: string | null) => eventId,
```
dqunbp, line 13: It is better to create the new action e.g. `deselectCurrentEvent`

vkozio: I'd not implement extra actions for such trivial things, because it will just add more "noise" for reatomv3 migration

src/features/events_list/atoms/autoSelectEvent.ts
albaranau, line 40: This seems like half fix for https://kontur.fibery.io/Tasks/Task/Select-a-top-disaster-after-switching-feed-on-DN2-17741/comment=22952

It selects top disaster from the new feed, but the warning toast still appears. This needs to be looked into.

It doesn't fix https://kontur.fibery.io/Tasks/Task/Deselect-a-disaster-on-DN2-after-using-map-view-filter-17740 at all

dqunbp:
This is wrong way

PR #680, opened 2024-04-02
fix: 18065: correct bivariate legend style calculation rules

IlyaLosikKontur:
https://kontur.fibery.io/Tasks/Task/Some-hexagons-from-the-building-completeness-layer-are-not-displayed-18065

## Summary by CodeRabbit

- **Refactor**
	- Improved logic for retrieving character and number data in bivariate legends.
	- Enhanced condition handling in bivariate style generation for better performance and accuracy.
	- Updated import statements and refined property types in the BivariateMatrixCell component for improved styling.

### DIFF
--- a/src/utils/bivariate/bivariateLegendUtils.ts
+++ b/src/utils/bivariate/bivariateLegendUtils.ts
@@ -89,6 +89,15 @@ export function createBivariateMeta(
   };
 }
 
+// need to ignore first range to avoid getting @ and 0 chars
+const indexTestFunc = <T extends { value: number }>(
+  el: T,
+  i: number,
+  arr: T[],
+  value: number,
+): boolean =>
+  (i !== 0 && value < el.value) || (i === arr.length - 1 && value <= el.value);
+
 // same logic as in classResolver method in styleGen.ts, but here we do it manually
 export function getCellLabelByValue(
   xSteps: Step[],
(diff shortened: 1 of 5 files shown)


src/utils/bivariate/bivariateLegendUtils.ts
```diff
@@ -89,19 +89,28 @@ export function createBivariateMeta(
   };
 }
 
+// need to ignore first range to avoid getting @ and 0 chars
+const indexTestFunc = <T extends { value: number }>(
```
dqunbp: Can we improve naming?

For example
```
const fitsLegendCriteria = <T extends { value: number }>(
  step: T,
  index: number,
  steps: T[],
  targetValue: number
): 
```

Do we need to pass the whole array, or steps count will be enough?

dqunbp:
Can we update the tests to handle the errors cases that will be fixed by the new implementation?

src/utils/bivariate/bivariate_style/styleGen.ts
```diff
@@ -66,6 +66,15 @@ function stringsToFeatureProp(expression) {
...
 export const getCharByIndex = (i: number) => String.fromCharCode(AT_CHAR_CODE + i); //get A - C by index
 
+/**
+ * Return condition based on border range
+ * all ranges - <
+ * last range - <=
+ * @param {number} i - index of current range
+ * @param {number} bL - length of borders array
+ */
+const getConditionFunc = (i: number, bL: number) => (i === bL - 1 ? lessOrEqual : less);
```
dqunbp: Can we improve the naming?

For example
```ts
const getConditionFunc = (currentIndex: number, totalBorders: number) =>
  currentIndex === totalBorders - 1 ? lessOrEqual : less;
```

dqunbp:
It looks like we don't need to include the task number from Fibery in the commit messages; we already have it in the branch name.

dqunbp (approved):

PR #689, opened 2024-04-09
fix(18180): add Record Sensors button to Toolbar

albaranau:
https://kontur.fibery.io/Tasks/Task/Live-sensor-button-is-not-present-on-Smart-City-app-18180

### DIFF
--- a/src/features/live_sensor/index.ts
+++ b/src/features/live_sensor/index.ts
@@ -11,7 +11,7 @@ export const liveSensorsControl = toolbar.setupControl<{
   id: SENSOR_CONTROL_ID,
   type: 'button',
   typeSettings: {
-    name: i18n.t('toolbar.start_sensor_recording'),
+    name: i18n.t('toolbar.record_sensors'),
     hint: {
       regular: i18n.t('live_sensor.start'),
       active: i18n.t('live_sensor.finish'),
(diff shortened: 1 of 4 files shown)


src/features/live_sensor/index.ts
```diff
@@ -2,23 +2,23 @@ import { notificationServiceInstance } from '~core/notificationServiceInstance';
...
 
 export const liveSensorsControl = toolbar.setupControl<{
   liveSensor?: LiveSensor;
 }>({
-  id: SENSOR_CONTROL,
+  id: SENSOR_CONTROL_ID,
   type: 'button',
   typeSettings: {
-    name: SENSOR_CONTROL_NAME,
+    name: i18n.t('toolbar.start_sensor_recording'),
```
Komzpa: it will become Stop once recording starts though, like the hint on the line below suggests?

Komzpa:
This thing used to be a table, not json. Is this an easy fix to make it a table again?
image

src/features/live_sensor/index.ts
albaranau: @Komzpa need to look deeper into this task to implement this, will discuss with Alesia later

albaranau: The hint doesn't seem to have a visual effect on the button. To implement dynamic text change I would need more time and learn more about the sensors recording functionality.

Discussed it with @alesiahil , decided to set a static "Record Sensors" text, for now.

albaranau: (P.S: sorry about the rebase, been experimenting)

Komzpa (approved):

PR #692, opened 2024-04-09
Refactor e2e test configuration and environment setup

dqunbp:
This commit introduces several changes to the end-to-end (e2e) testing setup and configuration. The primary changes include:

- Standardization of environment variable names across different files, replacing `TEST_EMAIL` and `TEST_PASSWORD` with `EMAIL` and `PASSWORD`.
- Removal of hardcoded environment URLs in the GitHub Actions workflow file, replacing them with a dynamic setup that adjusts based on the input environment.
- Deletion of an example test file that was no longer needed.
- Update of the login test specification to accommodate changes in the environment setup and to remove hardcoded values.
- Refactoring of the projects configuration JSON to include environment-specific URLs and additional properties such as `hasCookieBanner`.
- Adjustment of the Playwright configuration to prioritize the production environment file.

These changes aim to simplify the configuration process, reduce duplication, and make the e2e testing framework more flexible and maintainable.

### DIFF
--- a/e2e/login.spec.ts
+++ b/e2e/login.spec.ts
@@ -1,15 +1,34 @@
+import fs from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
 import { test, expect } from '@playwright/test';
-import projects from './projects-config.json' assert { type: 'json' };
+
+type Project = {
+  env: 'prod' | 'dev' | 'test';
+  name: string;
+  url: string;
+  title: string;
+  hasCookieBanner: boolean;
+};
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const data = fs.readFileSync(path.join(__dirname, 'projects-config.json')).toString();
+
+const appName = process.env.APP_NAME ?? 'all';
+const environment = process.env.ENVIRONMENT ?? 'prod';
+const projects: Project[] = JSON.parse(data)
+  .filter((project: Project) => project.env === environment)
+  .filter((project: Project) => appName === 'all' || project.name === appName);
 
 // Create a loop to loop over all the projects and create a test for everyone
-
-for (const project of Object.keys(projects)) {
-  const testProject = projects[project];
-  test(`As User, I can login to ${testProject.title}`, async ({ page }) => {
-    await page.goto(`${testProject.url}active/?app=${testProject.app}`);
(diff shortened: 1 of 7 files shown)


e2e/login.spec.ts
```diff
@@ -1,31 +1,49 @@
...
 
-for (const project of Object.keys(projects)) {
-  const testProject = projects[project];
-  test(`As User, I can login to ${testProject.title}`, async ({ page }) => {
-    await page.goto(`${testProject.url}active/?app=${testProject.app}`);
+const appName = process.env.APP_NAME ?? 'all';
+const environment = process.env.ENVIRONMENT ?? 'prod';
+const projects: Project[] = JSON.parse(data)
+  .filter((project: Project) => project.env === environment)
+  .filter((project: Project) => appName === 'all' || project.name === appName);
```
amurKontur, line 23: for what do we need 2 times filtering? Why not to do it in one iteration?

e2e/login.spec.ts
```diff
@@ -1,31 +1,49 @@
...
+  name: string;
+  url: string;
+  title: string;
+  hasCookieBanner: boolean;
+};
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const data = fs.readFileSync(path.join(__dirname, 'projects-config.json')).toString();
```
amurKontur, line 17: no so stable as tests could be grouped into folders in future, why not to use 'import projects from './projects-config.json' assert { type: 'json' }; '?

e2e/login.spec.ts
amurKontur, line 23: as it is about to be used for each test, better move it to a separate helper function ( this logics with projects gathering and getting config)

playwright.config.ts
```diff
@@ -5,7 +5,9 @@ import * as dotenv from 'dotenv';
  * https://github.com/motdotla/dotenv
  */
 
-dotenv.config({ path: ['.env.playwright.local', '.env.playwright'] });
+dotenv.config({
+  path: ['.env.playwright.production', '.env.playwright.local', '.env.playwright'],
```
amurKontur, line 9: for what do we need .env.playwright.production if we already have .env.playwright.local ? if we still need it then should we update readme ?

amurKontur (approved):

PR #713, opened 2024-04-22
fix(14214): Shift blue stroke of selected area outline to outside of the area shape

albaranau:
https://kontur.fibery.io/Tasks/Task/Blue-stroke-of-Selected-area-should-be-only-outside-of-event-shape-14214

## Summary by CodeRabbit

- **New Features**
  - Updated the geometry rendering to include a new line offset, enhancing the visual clarity of layered elements.

### DIFF
--- a/src/features/focused_geometry_layer/renderers/FocusedGeometryRenderer.ts
+++ b/src/features/focused_geometry_layer/renderers/FocusedGeometryRenderer.ts
@@ -32,6 +32,7 @@ const getLayersConfig = (
       paint: {
         'line-width': 8,
         'line-color': FOCUSED_GEOMETRY_COLOR,
+        'line-offset': -2,
       },
       layout: {
         'line-join': 'round',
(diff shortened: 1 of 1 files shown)


src/features/focused_geometry_layer/renderers/FocusedGeometryRenderer.ts
```diff
@@ -32,6 +32,7 @@ const getLayersConfig = (
       paint: {
         'line-width': 8,
         'line-color': FOCUSED_GEOMETRY_COLOR,
+        'line-offset': -2,
```
Komzpa: width 8 means that offset needs to be at least modulo 4 to be fully outside?

albaranau: Actually, should be -3, now that I double checked.
I tried -4 at first, but it looks a bit jumpy when switching Edit Mode on and off, so I used -2. 
-3 properly accounts for edit mode/disaster outline width (2)

albaranau: Updated to -3

albaranau: Updated again to move white outline outside the shape (this is a requirement that got lost previously and was just added to the ticket)

Komzpa (approved):

PR #718, opened 2024-04-29
[Localization] Create links check autotest ( complexed localization links vitest test )

amurKontur:
Newly added test that checks:

- Links are defined everywhere for all translations where set. 
- Texts describing links are defined everywhere for all translations where set
- Each link in english version is equal to translated link
- Each text describing link in english is not equal to text describing link in translated version (but present!)
- Each link block in english has corresponding translation if this translation is present
- Each link block in translation has corresponding english version if this translation is present

Checked and tested all this checks above being operable using corresponding input data.


src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  };
+  return result;
+};
+
+// Here is the test itself
+const compareLinksTest = (languageCode: string) => {
+  // First reading data from file
+  const poFileContent = fs.readFileSync(
+    `src/core/localization/gettext/${languageCode}/common.po`,
+  );
```
amurKontur, line 78: it is a static file located in the directory, so getting it in sync way does not block the call stack for too long.

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,85 @@
+import { describe, expect, it, test } from 'vitest';
+import gettextParser from 'gettext-parser';
+import fs from 'fs';
+
+// Here the input data for the test is specified
+const languageCodes = ['ar', 'de', 'es', 'id', 'ko', 'uk'];
+
+const processLinkAndText = (linkAndText) => {
+  const devidedArrLinkAndText = linkAndText.split(']');
```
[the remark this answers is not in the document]
amurKontur: fixed

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  const textOfLink = devidedArrLinkAndText[0].replace('[', '');
+  const link = devidedArrLinkAndText[1].slice(1, devidedArrLinkAndText[1].length - 1);
+  const result = {
+    link: link,
+    text: textOfLink,
+  };
+  return result;
+};
```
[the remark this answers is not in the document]
amurKontur: fixed

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    `src/core/localization/gettext/${languageCode}/common.po`,
+  );
+  // Using parser to parse data
+  const poData = gettextParser.po.parse(poFileContent).translations[''];
+  // Looping over converted object to array
+  for (const [reference, info] of Object.entries(poData)) {
+    // exclude "" key
+    if (reference) {
+      // Match [sometext](sometext)
+      const regularExpToFind = /\[.*?\]\(.*?\)/g;
```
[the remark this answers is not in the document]
amurKontur: fixed

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,84 @@
+import { describe, expect, it, test } from 'vitest';
+import gettextParser from 'gettext-parser';
+import fs from 'fs';
+
+// Here the input data for the test is specified
```
albaranau: No need to for this many detailed comments

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+const processLinkAndText = (linkAndText) => {
+  const [textOfLink, link] = linkAndText.split(']');
+  return {
+    link: link.slice(1, -1),
+    text: textOfLink.slice(1),
+  };
+};
+
+// Here is the test itself
```
albaranau: This comment isn't needed

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,84 @@
...
+const compareLinksTest = (languageCode: string) => {
+  // First reading data from file
```
albaranau: This comment isn't needed, it's obvious from the code

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  const poFileContent = fs.readFileSync(
+    `src/core/localization/gettext/${languageCode}/common.po`,
+  );
+  // Using parser to parse data
```
albaranau: This comment isn't needed, it's obvious from the code

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  const poData = gettextParser.po.parse(poFileContent).translations[''];
+
+  // Match [sometext](sometext)
```
albaranau: This comment isn't needed, it's obvious from the code

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,84 @@
...
+          const engTextAndLink = engVersionLinksArr?.[index];
+          // if translated version text with link is present, english version for it should also be present
+          expect(engTextAndLink).toBeTruthy();
+        });
+      }
+    }
+  }
+};
+
+// Building vitest test
```
albaranau: This comment isn't needed, it's obvious from the code

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,84 @@
...
+describe('Localisation links', () => {
```
albaranau: Localization

package.json
```diff
@@ -197,16 +198,17 @@
     "vite-plugin-html": "^3.2.2",
     "vite-plugin-mkcert": "^1.17.4",
     "vite-tsconfig-paths": "^4.3.2",
-    "vitest": "^1.3.1",
     "wretch": "^2.8.1"
   },
   "devDependencies": {
     "@playwright/test": "^1.42.1",
+    "@types/gettext-parser": "^4.0.4",
```
albaranau, line 208: Please also commit your pnpm-lock.yaml
It's out of sync with package.json in your branch, which causes errors in pipeline

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  const poFileContent = fs.readFileSync(
+    `src/core/localization/gettext/${languageCode}/common.po`,
+  );
+  // Using parser to parse data
+  const poData = gettextParser.po.parse(poFileContent).translations[''];
+
+  // Match [sometext](sometext)
+  const regularExpToFind = /\[.*?\]\(.*?\)/g;
+
+  // Looping over converted object to array
```
albaranau: It's a bit hard to understand this comment, but also there's probably no need to comment the loop

src/core/localization/linksNoTranslation.test.ts
albaranau: The name is a bit misleading. The test is about links in localized texts, but links aren't localized. 
"Links in localized texts" would be a bit better

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      const engVersion = info.msgid;
+      const translatedVersion = info.msgstr[0];
+
+      const engVersionLinksArr = engVersion.match(regularExpToFind);
+      const translatedVersionLinksArr = translatedVersion.match(regularExpToFind);
+
+      if (translatedVersion) {
+        // match() returns an array, looping it over considering that undefined might be as input if no links are present
+        engVersionLinksArr?.forEach((engTextAndLinkText, index) => {
+          const translatedTextAndLinkText = translatedVersionLinksArr?.[index];
```
albaranau: The whole body of this loop should probably be a separate function

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,84 @@
...
+          expect(engTextAndLink).toBeTruthy();
+        });
+      }
+    }
+  }
+};
+
+// Building vitest test
+describe('Localisation links', () => {
+  describe('Translated links are the same as original ones, but texts are different', () => {
```
albaranau: That's not the only check your'e asserting in your tests.

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,84 @@
+import { describe, expect, it, test } from 'vitest';
+import gettextParser from 'gettext-parser';
+import fs from 'fs';
+
+// Here the input data for the test is specified
+const languageCodes = ['ar', 'de', 'es', 'id', 'ko', 'uk'];
+
+const processLinkAndText = (linkAndText) => {
```
albaranau: Better name would be something like "parseLinkAndText" or "extractLinkAndText". The word "process" isn't really describing what the function is doing.

But also - you probably shouldn't be doing the parsing by hand, there're libraries for markdown parsing, for example:
[mdast-util-from-markdown](https://www.npmjs.com/package/mdast-util-from-markdown)

src/core/localization/linksNoTranslation.test.ts
albaranau: I don't think we should assume that translated text is always different from the English one.
"Info" -> "Info" is a legitimate translation case

albaranau (commented):
This is partial review, there may be more comments later

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,84 @@
...
+const compareLinksTest = (languageCode: string) => {
+  // First reading data from file
+  const poFileContent = fs.readFileSync(
+    `src/core/localization/gettext/${languageCode}/common.po`,
+  );
+  // Using parser to parse data
+  const poData = gettextParser.po.parse(poFileContent).translations[''];
+
+  // Match [sometext](sometext)
+  const regularExpToFind = /\[.*?\]\(.*?\)/g;
```
albaranau: You probably shouldn't be parsing Markdown by hand, there're libraries for it, for example:
[mdast-util-from-markdown](https://www.npmjs.com/package/mdast-util-from-markdown)

albaranau: https://gist.github.com/albaranau/aa7fd21782ba629cec8f9d3d81ee3823

Here's my example of using a library to parse links from a Markdown string. Feel free to use it

src/core/localization/linksNoTranslation.test.ts
amurKontur: deleted

src/core/localization/linksNoTranslation.test.ts
amurKontur: deleted

src/core/localization/linksNoTranslation.test.ts
amurKontur: deleted

src/core/localization/linksNoTranslation.test.ts
amurKontur: deleted

src/core/localization/linksNoTranslation.test.ts
amurKontur: deleted

src/core/localization/linksNoTranslation.test.ts
amurKontur: deleted

src/core/localization/linksNoTranslation.test.ts
amurKontur: renamed and also tests names improved

src/core/localization/linksNoTranslation.test.ts
amurKontur: deleted

src/core/localization/linksNoTranslation.test.ts
amurKontur: moved out

src/core/localization/linksNoTranslation.test.ts
amurKontur: fixed and added more logics in names

src/core/localization/linksNoTranslation.test.ts
amurKontur: imported and used library in a separate function

src/core/localization/linksNoTranslation.test.ts
amurKontur: added as separate function and imported to the test file. This logics might be useful somewhere in the code too

src/core/localization/extractLinkAndLabelFromMarkdown.ts
```diff
@@ -0,0 +1,34 @@
...
+        const label =
+          linkChild?.children[0]?.type === 'text'
+            ? linkChild.children[0].value
+            : undefined;
+        result.push({ link, label });
+      }
+    }
+  }
+  return result;
+};
```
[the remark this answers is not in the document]
amurKontur: it is impossible to have as the files worked through are generated by other library

src/core/localization/extractLinkAndLabelFromMarkdown.ts
```diff
@@ -0,0 +1,34 @@
...
+  link?: string;
+  label?: string;
+};
+
+export const extractLinkAndLabels = (markdownString: string): LinkAndLabel[] => {
+  const result: LinkAndLabel[] = [];
+  const paragraph = fromMarkdown(markdownString).children[0];
+  if (paragraph?.type === 'paragraph') {
+    const linkChildren: Link[] = paragraph.children.filter(
+      (chart) => chart.type === 'link',
```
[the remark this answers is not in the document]
amurKontur: it is from the word charter

package.json
amurKontur: done
there were many other problems but I fixed them

src/core/localization/extractLinkAndLabelFromMarkdown.ts
```diff
@@ -0,0 +1,34 @@
...
+    ) as Link[];
+    for (const linkChild of linkChildren) {
+      const link = linkChild?.url;
+      if (link) {
+        const label =
+          linkChild?.children[0]?.type === 'text'
+            ? linkChild.children[0].value
+            : undefined;
```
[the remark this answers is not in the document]
amurKontur: no, you break KISS principle by this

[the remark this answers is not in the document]
amurKontur: and its okey to have it like that

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+      if (translatedVersion) {
+        checkLinksAndLabels(engVersionLinksArr, translatedVersionLinksArr);
+      }
+    }
+  }
+};
+
+describe('Links in localized texts', () => {
+  describe('Links are untranslated, have corresponding label, and are properly defined in links and labels of translations', () => {
```
albaranau: It's a bit hard to understand what "and are properly defined in links and labels of translations" means. Very long test name overall

src/core/localization/extractLinkAndLabelFromMarkdown.ts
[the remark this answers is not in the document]
albaranau: What coderabbit proposed technically would be okay (and better) if our goal was to find the first child with "text" type. In this case it's assumed there will only be one root child because it's a link type.

src/core/localization/extractLinkAndLabelFromMarkdown.ts
albaranau, line 1: I don't see a reason to keep this function in a separate file, it reduces coverage and adds clutter to folder. Better keep it inside test file

src/core/localization/extractLinkAndLabelFromMarkdown.ts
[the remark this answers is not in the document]
albaranau: What "charter" means in this context? :) I agree with coderabbit on this one,"child" makes more sense, since we're filtering children values array

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  [key: string]: GettextTranslation;
+};
+
+const checkLinksAndLabels = function (
+  engArray: LinkAndLabel[],
+  translatedArray: LinkAndLabel[],
+) {
+  engArray.forEach((engLinkAndLabel: LinkAndLabel, index) => {
+    const translatedLinkAndLabel: LinkAndLabel = translatedArray[index];
+    // If eng version text with link is present and text in translated version is present, verify that translated version has text with link under the same number and generally has text with link.
```
albaranau: This comment is too long (I had to use horizontal scroll), would be better to break it down in multiple lines or make it shorter

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  });
+};
+
+const compareLinksTest = (languageCode: string) => {
+  const poFileContent = fs.readFileSync(
+    `src/core/localization/gettext/${languageCode}/common.po`,
+  );
+
+  const poDataFull = gettextParser.po.parse(poFileContent);
+  const poData = poDataFull.translations[''] as GettextTranslations;
```
albaranau: You can replace `GettextTranslations` with `Record<string, GettextTranslation>`
(Record is a convenience type for `[key: K]: T`)

GettextTranslations type can be removed then.
It will be easier to read and more compact.

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  for (const [reference, info] of Object.entries(poData) as [
+    string,
+    GettextTranslation,
+  ][]) {
+    // exclude "" key
+    if (reference) {
+      const engVersion = info.msgid;
+      const translatedVersion = info.msgstr[0];
+
+      const engVersionLinksArr = extractLinkAndLabels(engVersion);
```
albaranau: NIT: It would be easier to read as `engLinksArr` or `englishLinksArr`.
(Same with `translatedLinksArr`)

src/core/localization/linksNoTranslation.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  const poDataFull = gettextParser.po.parse(poFileContent);
+  const poData = poDataFull.translations[''] as GettextTranslations;
+
+  for (const [reference, info] of Object.entries(poData) as [
+    string,
+    GettextTranslation,
+  ][]) {
+    // exclude "" key
+    if (reference) {
+      const engVersion = info.msgid;
```
albaranau: NIT: it would be easier for understanding to have `engText` and `translatedText` instead of `engVersion/translatedVersion`

src/core/localization/extractLinkAndLabelFromMarkdown.ts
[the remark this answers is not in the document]
albaranau: While technically possible to wrap parsing in try-catch, this function will only be used in test, so it should be okay to leave as is

albaranau (changes requested):

src/core/localization/linksNoTranslation.test.ts
albaranau: Please move the function inside the test file. It's unlikely it will be used somewhere else, and extra file clutters up the space.
If we really wanted to use it, it would be better to move it to some of the utils folders.

src/core/localization/linksNoTranslation.test.ts
albaranau: "Links must be identical between different languages and must all have labels"

src/core/localization/linksNoTranslation.test.ts
amurKontur: done

src/core/localization/extractLinkAndLabelFromMarkdown.ts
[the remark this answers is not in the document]
amurKontur: done

src/core/localization/linksNoTranslation.test.ts
amurKontur: done

src/core/localization/extractLinkAndLabelFromMarkdown.ts
amurKontur: done

src/core/localization/linksNoTranslation.test.ts
amurKontur: fixed

src/core/localization/linksNoTranslation.test.ts
amurKontur: fixed

src/core/localization/linksNoTranslation.test.ts
amurKontur: fixed

src/core/localization/linksNoTranslation.test.ts
amurKontur: fixed

albaranau (approved):

PR #724, opened 2024-05-10
fix(18441): redirect to Atlas login after registration in Keycloak

alesiahil:
## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the redirect URLs in the Login Form to ensure users are directed to the correct profile page after registration.


src/features/user_profile/components/LoginForm/LoginForm.tsx
```diff
@@ -11,7 +11,7 @@ import type { ChangeEvent } from 'react';
 
 const authInputClasses = { input: clsx(s.authInput) };
 
-const registrationUrlEncoded = encodeURIComponent(`${configRepo.get().keycloakUrl}/realms/${configRepo.get().keycloakRealm}/protocol/openid-connect/registrations?client_id=account&response_type=code&scope=openid&redirect_uri=https://www.kontur.io/portfolio/`);
+const registrationUrlEncoded = encodeURIComponent(`${configRepo.get().keycloakUrl}/realms/${configRepo.get().keycloakRealm}/protocol/openid-connect/registrations?client_id=account&response_type=code&scope=openid&redirect_uri=https://atlas.kontur.io/active/profile`);
```
albaranau: This will redirect to Atlas from any of the applications. I don't think we want that.
FE team should pick up this task, there's a bit more to do than simply replace the URL

albaranau: Apparently, I missed the discussion about us planning to redirect to Atlas from any app.
In that case, this looks okay :)

albaranau (approved):

PR #726, opened 2024-05-13
feat(18419): Beautify MCDA layer analysis editing

Natallia-Harshunova:
https://kontur.fibery.io/Tasks/My-tasks-3575#Task/Beautify-MCDA-layer-analysis-editing-18419


src/components/LayerControl/LayerControl.module.css
```diff
@@ -33,6 +33,7 @@
   justify-content: center;
   gap: var(--unit);
   user-select: none;
+  margin-right: var(--half-unit);
```
albaranau, line 36: There's no need to add padding to these controls right now, we're only interested in MCDA layer editing in this ticket

albaranau:
Could you also make the Range and Weight fields bigger, please? 
image

Similarly to how it's shown in Figma, expect maybe Weight the same size as a single Range field:
image
https://www.figma.com/file/cFmWZmGioB52BDMoCofzpY/MCDA?type=design&node-id=16620-7697&mode=design&t=PyG3HuBQihpRdPX8-0

src/components/LayerControl/LayerControl.module.css
albaranau: ..on the other hand, it fixes alignment with Deselect button texts. Let it be :)

albaranau:
The font of range text input is now bigger than the rest of the inputs. Could you please fix this? 
image

Must be because `rangeInputBox` css class doesn't have the same rule for fontSize that `textInputBox` has.

albaranau (approved):

albaranau (approved):

PR #728, opened 2024-05-14
adding Atlas About page text

atsiatserkina:
at the end of the file strings of Atlas About page were added for translation in POT format


albaranau:
Since text (with its translated versions) will be stored in assets and not on FE, it would make more sense to store it in related repository (e.g. User Profile Service)

albaranau:
Since text (with its translated versions) will be stored in assets and not on FE, it would make more sense to store it in related repository (e.g. User Profile Service)

PR #737, opened 2024-05-22
feat(18186): add emoji to MCDA layers

Natallia-Harshunova:
https://kontur.fibery.io/Tasks/My-tasks-3575#Task/add-indicators-emojis-support-to-DN2-FE-18186

### DIFF
--- a/src/features/mcda/utils/generateEmojiPrefix.ts
+++ b/src/features/mcda/utils/generateEmojiPrefix.ts
@@ -1,11 +1,15 @@
 import emojiRegex from 'emoji-regex';
 
-export const generateEmojiPrefix = (emoji = '') => {
+const SPACE_CHAR = '\u3000';
+
+export const generateEmojiPrefix = (emoji = '', targetLength = 2) => {
   try {
     const regex = emojiRegex();
     const emojiCount = [...emoji.matchAll(regex)].length;
-    return emojiCount === 1 ? `\u3000\u00A0\u00A0${emoji}` : `${emoji}`;
+    return emojiCount < targetLength
+      ? SPACE_CHAR.repeat(targetLength - emojiCount) + emoji
+      : emoji;
   } catch {
-    return '';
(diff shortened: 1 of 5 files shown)


src/features/mcda/utils/generateEmojiPrefix.ts
```diff
@@ -0,0 +1,11 @@
+import emojiRegex from 'emoji-regex';
+
+export const generateEmojiPrefix = (emoji = '') => {
+  try {
+    const regex = emojiRegex();
+    const emojiCount = [...emoji.matchAll(regex)].length;
+    return emojiCount === 1 ? `\u3000\u00A0\u00A0${emoji}` : `${emoji}`;
```
albaranau: Why the combination of two different kinds of symbols here?
Current solution doesn't seem to be working as expected:
image

src/features/mcda/utils/generateEmojiPrefix.ts
```diff
@@ -0,0 +1,11 @@
...
+  } catch {
+    return '';
```
albaranau: Why not just return original `emoji` in this case?

src/features/mcda/utils/generateEmojiPrefix.ts
albaranau: Something like this seems to work:

```
export const generateEmojiPrefix = (emoji = '', targetLength = 2) => {
  try {
    const regex = emojiRegex();
    const emojiCount = [...emoji.matchAll(regex)].length;
    return emojiCount < targetLength
      ? '\u3000'.repeat(targetLength - emojiCount) + emoji
      : emoji;
  } catch {
    return emoji;
  }
};
```

albaranau: UPD: it seems like the emoji width isn't always consistent across different platforms and fonts

albaranau (approved):

PR #742, opened 2024-05-27
feat: disable unused params from location url

dqunbp:
hide the `feed` param if the app won't use it
hide the `app` param for the urls with custom domain

### DIFF
--- a/src/core/url_store/atoms/urlStore.ts
+++ b/src/core/url_store/atoms/urlStore.ts
@@ -115,7 +115,8 @@ export const urlStoreAtom = createAtom(
     newState.event = currentEvent?.id ? currentEvent.id : undefined;
 
     const currentFeed = get('currentEventFeedAtom');
-    newState.feed = currentFeed && isFeedSelectorEnabled ? currentFeed.id : undefined;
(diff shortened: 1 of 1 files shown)


src/core/url_store/atoms/urlStore.ts
```diff
@@ -31,7 +32,16 @@ export const urlStoreAtom = createAtom(
...
     init: (initialState: UrlData) => initialState,
   },
-  ({ get, schedule, onAction, onInit, create }, state: UrlData | null = null) => {
+  (
+    { get, schedule, onAction, onInit, create, getUnlistedState },
+    state: UrlData | null = null,
+  ) => {
+    const isFeedSelectorEnabled =
+      typeof (
+        getUnlistedState(featureFlagsAtom)[FeatureFlag.EVENTS_LIST__FEED_SELECTOR] ||
```
albaranau: From what @vkozio told me, it's probably better to use `configRepo.get().features` instead of `featureFlagsAtom`, because it's planned to get rid of `featureFlagsAtom` in the future

albaranau (approved):

PR #764, opened 2024-06-10
fix(18079): Add name parameter to MCDA configuration. Refactor opening mcda json file

albaranau:
https://kontur.fibery.io/Tasks/Task/Refactor-new-MCDA-name-and-id-creation-18079

## Summary by CodeRabbit

- **New Features**
  - Added a new error message for incorrect MCDA version.
  - Introduced a `name` property for MCDA configurations.
  - Added a default MCDA name constant: 'MCDA layer'.
  - Implemented a new function to generate MCDA IDs based on names.
  - Enhanced JSON file handling for MCDA configurations with improved error handling and validation.

### DIFF
--- a/src/features/mcda/utils/openMcdaFile.ts
+++ b/src/features/mcda/utils/openMcdaFile.ts
@@ -1,6 +1,7 @@
 import { i18n } from '~core/localization';
 import { currentNotificationAtom } from '~core/shared_state';
-import { isMCDAConfig } from '../mcdaConfig';
+import { DEFAULT_MCDA_NAME } from '../constants';
+import { generateMCDAId } from './generateMCDAId';
 import type { MCDAConfig } from '~core/logical_layers/renderers/stylesConfigs/mcda/types';
 
 const input = (() => {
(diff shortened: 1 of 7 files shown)


src/features/mcda/utils/openMcdaFile.ts
```diff
@@ -10,6 +11,29 @@ const input = (() => {
...
+  const result: Partial<MCDAConfig> = { ...json };
+  if (!result.version || result.version < 4) {
+    throw new Error(i18n.t('mcda.error_wrong_mcda_version'));
+  }
+  if (
+    !result?.layers?.length ||
+    !(
+      result?.colors?.type === 'sentiments' ||
+      result?.colors?.type === 'mapLibreExpression'
+    )
```
Natallia-Harshunova, line 24: Can be simplified to smth like this 
`!['sentiments', 'mapLibreExpression'].includes(result?.colors?.type)`

Natallia-Harshunova (approved):

src/features/mcda/utils/openMcdaFile.ts
```diff
@@ -10,6 +11,29 @@ const input = (() => {
   return input;
 })();
 
+function createMCDAConfigFromJSON(json: Partial<MCDAConfig>): MCDAConfig {
+  const result: Partial<MCDAConfig> = { ...json };
+  if (!result.version || result.version < 4) {
```
Natallia-Harshunova, line 16: Is it ok to use here magic number `4`?

albaranau: Should be fine, the version is literally defined like this in interface right now:
image
This check is pretty arbitraty tbh.

When new version rolls out, we might want to reorganize it better (we may want backwards compatibility and handle files of different versions differently)

src/features/mcda/utils/openMcdaFile.ts
albaranau: image
This simplification doesn't exactly work without additional checks or adding empty string just for the sake of having correct type. I will leave it as is for now

src/features/mcda/utils/openMcdaFile.ts
```diff
@@ -10,6 +11,29 @@ const input = (() => {
...
+    throw new Error(i18n.t('mcda.error_invalid_file'));
+  }
+  if (!result.id) {
+    result.id = generateMCDAId(result.name);
+  }
+  if (!result.name) {
+    result.name = result.id ?? DEFAULT_MCDA_NAME;
+  }
+  return result as MCDAConfig;
+}
```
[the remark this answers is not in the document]
albaranau: Please never again propose changing regular function to arrow function just "for consistency and to reduce complexity". This is not neeeded.

PR #766, opened 2024-06-11
fix(18607): fix MCDA style: paint faulty values (null, NaN, etc.) as transparent

albaranau:
https://kontur.fibery.io/favorites/Tasks/My-tasks-3575#Tasks/Task/Nulls-in-MCDA-layer-are-red-instead-of-transparent-18607

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved color mapping logic for sentiment analysis results, ensuring better handling of incorrect values (`null`, `NaN`) with a default transparent color.

### DIFF
--- a/src/core/logical_layers/renderers/stylesConfigs/mcda/mcdaStyle.ts
+++ b/src/core/logical_layers/renderers/stylesConfigs/mcda/mcdaStyle.ts
@@ -85,7 +85,7 @@ function sentimentPaint({
     'fill-color': [
       'let',
       'mcdaResult',
-      ['to-number', mcdaResult, -9999], // falsy values become -9999
+      mcdaResult,
       [
         'case',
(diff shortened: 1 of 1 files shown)


src/core/logical_layers/renderers/stylesConfigs/mcda/mcdaStyle.ts
```diff
@@ -85,7 +85,7 @@ function sentimentPaint({
     'fill-color': [
       'let',
       'mcdaResult',
-      ['to-number', mcdaResult, -9999], // falsy values become -9999
```
albaranau, line 88: Old code was converting any faulty value (like NaN or null) into -9999 which caused them to be colored instead of being transparent

dqunbp (approved):

PR #767, opened 2024-06-12
fix(mcdaStyle): 18787 fix style calculation for MCDA with a single axis

albaranau:
https://kontur.fibery.io/Tasks/Task/Weight-works-incorrectly-for-MCDA-with-a-single-axis-18787

## Summary by CodeRabbit

- **Refactor**
  - Simplified the logic of the normalization function for better performance and readability.
  - Enhanced the handling of default color values for incorrect inputs to ensure more consistent rendering.

### DIFF
--- a/src/core/logical_layers/renderers/stylesConfigs/mcda/mcdaStyle.ts
+++ b/src/core/logical_layers/renderers/stylesConfigs/mcda/mcdaStyle.ts
@@ -49,9 +49,8 @@ export function filterSetup(layers: MCDAConfig['layers']) {
 }
 
 export function linearNormalization(layers: MCDAConfig['layers']) {
-  const layersCount = layers.length;
-  if (layersCount === 1) {
-    return calculateLayer(layers.at(0)!);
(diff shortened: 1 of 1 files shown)


src/core/logical_layers/renderers/stylesConfigs/mcda/mcdaStyle.ts
```diff
@@ -49,9 +49,8 @@ export function filterSetup(layers: MCDAConfig['layers']) {
 }
 
 export function linearNormalization(layers: MCDAConfig['layers']) {
-  const layersCount = layers.length;
-  if (layersCount === 1) {
-    return calculateLayer(layers.at(0)!);
```
albaranau, line 54: Old code was not dividing the result by coefficient (weight), which is a calculation error. 

MCDA formula is:
`V_total = Σ(V_norm(1) * weight(1), V_norm(2) * weight(2), V_norm(n) * weight(n)) / Σweight`

dqunbp (approved):

PR #771, opened 2024-06-20
feat: Refresh auth tokens during the app initialization

dqunbp:
The `refreshAuthToken` method in the `ApiClient` class was changed from private to public. This change allows external code to refresh the authentication token. This is necessary to get fresh roles data from Keycloak.

https://kontur.fibery.io/Tasks/Task/Refresh-access-token-during-Kontur-app-initialization-18527

## Summary by CodeRabbit

- **New Features**
  - Improved authentication handling to ensure roles data is consistently refreshed when the user is authenticated.

### DIFF
--- a/src/core/api_client/apiClient.ts
+++ b/src/core/api_client/apiClient.ts
@@ -193,7 +193,7 @@ export class ApiClient {
   /**
    * @throws {ApiClientError}
    */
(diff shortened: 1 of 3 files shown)


albaranau (approved):

albaranau:
Let's rename the PR so it describes the change better

PR #799, opened 2024-07-17
feat(osm_edit_link): set default osmEditor value to josm

dqunbp:
https://kontur.fibery.io/Tasks/Task/BE-Default-OSM-editor-usage-causes-Unknown-editor-error-18784

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the logic for obtaining the `osmEditor` value from the store, providing a default value of `'josm'` to prevent errors when the value is not present.

### DIFF
--- a/src/features/osm_edit_link/index.tsx
+++ b/src/features/osm_edit_link/index.tsx
@@ -22,8 +22,7 @@ osmEditControl.onStateChange((ctx, state) => {
     try {
       const position = store.getState(currentMapPositionAtom);
       if (!position) throw Error('Unknown position');
-      const { osmEditor } = store.getState(currentUserAtom);
-      if (!osmEditor) throw Error('Unknown editor');
+      const { osmEditor = 'josm' } = store.getState(currentUserAtom);
 
       const editor = configRepo
         .get()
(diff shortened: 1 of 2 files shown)


src/features/osm_edit_link/index.tsx
```diff
@@ -22,8 +22,7 @@ osmEditControl.onStateChange((ctx, state) => {
     try {
       const position = store.getState(currentMapPositionAtom);
       if (!position) throw Error('Unknown position');
-      const { osmEditor } = store.getState(currentUserAtom);
-      if (!osmEditor) throw Error('Unknown editor');
+      const { osmEditor = 'josm' } = store.getState(currentUserAtom);
```
albaranau: Let's make a constant for DEFAULT_OSM_EDITOR maybe?

albaranau (approved):

src/features/osm_edit_link/index.tsx
Komzpa: Notes about this:
  - default should be "OpenStreetMap.org default editor" which will be configured on osm.org site
  - that option was lost on some refactoring, essentially it is similar to what iD link is but without specifying the editor explicitly "https://www.openstreetmap.org/edit#map="
  - while at it, the link to the iD editor has unneeded node=.... parameter that should be deleted

PR #819, opened 2024-08-08
fix(core): 18520 auth fixes and refactors

vkozio:
- implemented end session flow via endpoint

- automatic logout only on 401 from dedicated request to token endpoint

- refresh token during authClient init, all requests will use fresh token from app start

Refactor:

- extracted token management from apiClient to OidcSimpleClient,

- authClientInstance is OidcSimpleClient now, AuthClient and irrelevant code removed, necessary methods are used directly

- adjusted naming in auth-related code to align more with standarts and existing solutions

- move ~core/auth/types to ~core/app/types as it has app-related types

- simplify and streamline init flow, remove unecessary atoms dependencies, reduce early activation issues

- reduce reactive usage of featureFlags

Addidtional fixes:

- add Suspense with preloader in view hierarchy to keep sidebar visible during loading, fixes (19141) https://kontur.fibery.io/Tasks/Task/White-screen-fails-while-loading-application-19141

- map adapter: removed heavy react-dom/server dependency (~50kb in bundle) and unused or irrelevant code

- drop date-fns in favor of native function

## Summary by CodeRabbit

- **New Features**
  - Introduced warnings for unused variables and expressions to encourage cleaner code practices.
  - Updated various dependencies to enhance performance and security.
  - Enhanced authentication mechanism for a smoother user experience during API interactions.
  - Refactored link handling in the Markdown component for improved modularity and clarity.

- **Bug Fixes**
  - Improved handling of unauthorized responses to minimize disruptions due to token expiration.


src/boot.ts
```diff
@@ -19,28 +20,25 @@ export async function setupApplicationEnv() {
...
   const sharedConfig = readInitialUrl();
 
-  // TODO - initialUrl from config repository
-  localStorage.setItem('initialUrl', location.href); // keep initial url before overwriting by router
-
   // Stage config - depends from appconfig.json in CI repo
   // (/configs/config.local.json for localhost)
   const stageConfig = await getStageConfig();
 
+  authClientInstance.init(
```
Natallia-Harshunova, line 27: Don't we need to `await `this function call as it async?

Natallia-Harshunova (approved):

package.json
```diff
@@ -94,15 +94,16 @@
...
     "@reatom/logger": "^3.8.4",
     "@reatom/npm-react": "^3.8.10",
     "@reatom/react-v2": "^3.1.2",
-    "@sentry/react": "^7.113.0",
-    "@slack/web-api": "^7.0.4",
-    "@testing-library/react": "^14.2.1",
+    "@sentry/react": "^8.18.0",
+    "@slack/web-api": "^7.3.1",
+    "@testing-library/dom": "^10.3.2",
+    "@testing-library/react": "^16.0.0",
```
dqunbp, line 106: Isn't it better to put testing libraries in `dev` dependencies?

src/core/apiClientInstance.ts
```diff
@@ -25,7 +15,7 @@ export const apiClient = withSetupCheck(ApiClient, {
   },
 });
 
-export const reportsClient = withSetupCheck(ApiClient, {
+export const reportsClient = new ApiClient({
```
dqunbp, line 18: It looks like after all the changes `apiClient` and `reportsClient` have exactly the same code. Do we need both separately?

package.json
vkozio: Yes. Actually it's a good question to those, who added them initially in main deps)

vkozio: to do in #19349

src/boot.ts
vkozio: Good catch) It initiates token refresh in "background" and we don't need to wait here

src/core/apiClientInstance.ts
vkozio: It's different endpoints, using different baseUrls, that's why we have separate Clients

PR #857, opened 2024-10-05
fix(core): 18906 redirect to plans page after login on atlas

vkozio:
## Release Notes

- **New Features**
	- Enhanced routing logic to redirect users based on authentication status and feature availability.
	- Added support for dynamic route IDs in the `CommonView` component.

- **Improvements**
	- Refined CSS styles for article components to improve layout and presentation.
	- Introduced an optional `id` property in the `Article` component for greater flexibility.

- **Bug Fixes**
	- Updated ignored file patterns to streamline cursor operations.

- **Documentation**
	- Clarified routing constants for better accessibility across modules.


src/core/router/routes.tsx
```diff
@@ -24,10 +24,11 @@ const { BivariateManagerPage } = lazily(
   () => import('~views/BivariateManager/BivariateManager'),
 );
 
-const isAuthenticated = !!configRepo.get().user;
+export const isAuthenticated = !!configRepo.get().user;
+export const isMapFeatureEnabled = configRepo.get().features[AppFeature.MAP];
```
dqunbp, line 28: Do these values update reactively?  
Are these constants related to routes?

dqunbp: It seems that these constants are meaningful for the application as a whole and not just for routes.

dqunbp (approved):

PR #868, opened 2024-10-17
feat(about): 19871 Add "User Guide" page

albaranau:
https://kontur.fibery.io/Tasks/Task/add-User-Guide-subtab-to-About-page-19871
https://kontur.fibery.io/Tasks/Task/Display-About-subpages-on-Kontur-app-sidebar-according-to-about_page-config-18590

image

## Summary by CodeRabbit

- **New Features**
	- Introduced a user guide sub-tab under the about page for enhanced navigation.
	- Added a new entry for "User guide" in the localization settings to improve user interface clarity.
	- Enhanced routing configuration to dynamically include new routes based on feature settings.

- **Bug Fixes**
	- Improved functionality for inner anchor links to ensure they work correctly within the same page.

- **Documentation**
	- Updated mock configuration data to better simulate the application environment for testing and development.
	- Added new type definitions to extend existing configurations without altering functionality.


src/core/config/types.ts
```diff
@@ -90,3 +90,12 @@ export type FaviconPack = Record<
...
 >;
+
+export type AboutFeatureConfig = {
+  tabId: string;
+  assetUrl: string;
+  subTabs: {
+    tabId: string;
+    assetUrl: string;
+  }[];
+};
```
Natallia-Harshunova: Tecnically, we can make recursive types in that case. Field can have the same type as  whole object. But in this case there better be a different name for the type.
```
export type AboutFeatureConfig = {
  tabId: string;
  assetUrl: string;
  subTabs: AboutFeatureConfig[];
};
```

src/core/router/routes.tsx
```diff
@@ -27,6 +29,54 @@ const { BivariateManagerPage } = lazily(
...
+    icon: <BookOpen24 />,
+    visibilityInNavigation: 'always',
+    requiredFeature: AppFeature.ABOUT_PAGE,
+  },
+};
+
+function getAboutSubTabs() {
+  const subTabsConfig = configRepo?.get().features.about_page?.['subTabs'] as
+    | AboutFeatureConfig['subTabs']
+    | undefined;
```
Natallia-Harshunova:
```suggestion
  const subTabsConfig = configRepo.get().features[AppFeature.ABOUT_PAGE]?.['subTabs'] as
    | AboutFeatureConfig['subTabs']
    | undefined;
```

src/core/router/routes.tsx
```diff
@@ -27,6 +29,54 @@ const { BivariateManagerPage } = lazily(
...
+  if (Array.isArray(subTabsConfig)) {
+    return subTabsConfig
+      .filter((subTab) => ABOUT_SUBTABS[subTab.tabId] && subTab.assetUrl)
```
Natallia-Harshunova, line 65: If `assetUrl `optional then make appropriate changes to the type

Natallia-Harshunova (approved):

src/core/config/types.ts
albaranau: Mmm, I would probably make these types recursive once we actually start processing these objects recursively. Otherwise it just feels like unneeded complexity

src/core/router/routes.tsx
albaranau: Fixed

src/core/router/routes.tsx
albaranau: Fixed

PR #891, opened 2024-11-07
fix(map navigation): 18438 Speed up navigation to a point on the map

albaranau:
Removing extra layer of abstraction when navigating to a point on the map.

https://kontur.fibery.io/Tasks/Task/Speed-up-zoom-when-using-Locate-me-(Breadcrumbs,-search)-feature-18438
https://kontur.fibery.io/Tasks/Task/FE-Test1-Locate-me-feature-does-not-zoom-to-location-for-new-OAM-18508

## Summary by CodeRabbit

- **New Features**
	- Introduced a new action handler `updateCurrentMapPosition` for enhanced map position updates.
	- Added a custom hook `useMapPositionSync` for real-time synchronization of the map's position with shared state.
- **Improvements**
	- Simplified synchronization logic for map position and bounding box actions.
	- Renamed hook from `useMapPositionSmoothSync` to `useMapPositionSync` for clarity and improved functionality.
- **Bug Fixes**
	- Resolved potential issues with cyclic updates by refining the map position synchronization process.


albaranau:
@coderabbitai pause

albaranau:
PR not yet ready for review, WIP

src/core/shared_state/currentMapPosition.ts
```diff
@@ -1,70 +1,77 @@
...
 
 export type CurrentMapPositionAtomState = MapPosition | null;
 
+// TODO: #20160 update currentMapPositionAtom to reatom v3
 export const currentMapPositionAtom = createAtom(
   {
     setCurrentMapPosition: (mapPosition: { lat: number; lng: number; zoom: number }) =>
       mapPosition,
     setCurrentMapBbox: (mapBbox: Bbox | [[number, number], [number, number]]) => mapBbox,
+    updateCurrentMapPosition: (mapPosition: { lat: number; lng: number; zoom: number }) =>
```
Natallia-Harshunova: I think we can define `mapPosition` type as `CenterZoomPosition` as well as  in `setCurrentMapPosition`

src/core/shared_state/currentMapPosition.ts
```diff
@@ -1,70 +1,77 @@
...
-        state = position;
-      } else {
-        const { lat, lng, zoom } = position;
-        if (
-          'lng' in state &&
-          (state.lat !== lat || state.lng !== lng || state.zoom !== zoom)
-        ) {
-          state = position;
+    function jumpTo(map: Map, position: MapPosition) {
+      const { lng, lat, zoom } = position as CenterZoomPosition;
```
Natallia-Harshunova:
```suggestion
    function jumpTo(map: Map, position: CenterZoomPosition) {
      const { lng, lat, zoom } = position;
```

src/core/shared_state/currentMapPosition.ts
```diff
@@ -1,70 +1,77 @@
...
-      if (state === null || !('lng' in state)) {
-        state = position;
-      } else {
-        const { lat, lng, zoom } = position;
-        if (
-          'lng' in state &&
-          (state.lat !== lat || state.lng !== lng || state.zoom !== zoom)
-        ) {
-          state = position;
+    function jumpTo(map: Map, position: MapPosition) {
```
Natallia-Harshunova: This function is independent from atom, you can move it out of it

src/core/shared_state/currentMapPosition.ts
```diff
@@ -1,70 +1,77 @@
...
 export const currentMapPositionAtom = createAtom(
   {
     setCurrentMapPosition: (mapPosition: { lat: number; lng: number; zoom: number }) =>
       mapPosition,
     setCurrentMapBbox: (mapBbox: Bbox | [[number, number], [number, number]]) => mapBbox,
+    updateCurrentMapPosition: (mapPosition: { lat: number; lng: number; zoom: number }) =>
+      mapPosition,
+    currentMapAtom,
   },
   ({ onAction }, state: CurrentMapPositionAtomState = null) => {
```
Natallia-Harshunova, line 47: Now state cannot be the type of `BboxPosition`, so `CurrentMapPositionAtomState ` can be simplified to `CurrentMapPositionAtomState = CenterZoomPosition | null`

src/core/shared_state/currentMapPosition.ts
```diff
@@ -1,70 +1,77 @@
...
 export type MapPosition = CenterZoomPosition | BboxPosition;
 
 export type CurrentMapPositionAtomState = MapPosition | null;
 
+// TODO: #20160 update currentMapPositionAtom to reatom v3
 export const currentMapPositionAtom = createAtom(
   {
     setCurrentMapPosition: (mapPosition: { lat: number; lng: number; zoom: number }) =>
       mapPosition,
     setCurrentMapBbox: (mapBbox: Bbox | [[number, number], [number, number]]) => mapBbox,
```
Natallia-Harshunova, line 43: `setCurrentMapBbox` doesn't return this type i guess

Natallia-Harshunova (commented):
Yes, It works much faster now! Cool
I just think that its better to update types for better migration to v3, i tried to point out the changes, there are some redundant and misleading ones

src/core/shared_state/currentMapPosition.ts
dqunbp: @albaranau  `CenterZoomPosition ` There is a simpler and shorter name for this object - `Viewport`

src/core/shared_state/currentMapPosition.ts
```diff
@@ -1,70 +1,77 @@
...
-      if (cam.center && 'lng' in cam.center) {
-        const { zoom } = cam;
-        const { lat, lng } = cam.center;
-        if (
-          prev == null ||
-          ('lng' in prev && (prev.lat !== lat || prev.lng !== lng || prev.zoom !== zoom))
-        ) {
-          state = { ...position, lat, lng, zoom };
+      if (map) {
+        const cam = getCameraForBbox(bbox, map);
```
dqunbp: Not `cam` but `viewport`

albaranau: @dqunbp  I can't find, where is the concept of viewport coming from here? 
`getCameraForBbox` uses Maplibre's `map.cameraForBounds()` and returns CenterZoomBearing object

I never heard about Center+Zoom+Bearing being called a Viewport before, sounds counterintuitive (doesn't Viewport usually describe the portion of the screen). Where can I read about it?

src/core/shared_state/currentMapPosition.ts
albaranau: There is a case when state can still need `BboxPosition` - when bbox is passed as a URL parameter, on startup - map isn't initialized yet, but bbox coordinates need to be stored.

I would rather not try cutting out any functionality in the scope of this task.

src/core/shared_state/currentMapPosition.ts
albaranau: It's not a return type, though it does look like it, it's just how reatom v2 declares action parameters. It is pretty confusing

src/core/shared_state/currentMapPosition.ts
albaranau: Seems like it's better not to rename the type, because actual Viewport is expected to have more info:
https://konturio.slack.com/archives/C02DEKP2MKP/p1731932373198899?thread_ts=1731931161.995599&cid=C02DEKP2MKP

src/core/shared_state/currentMapPosition.ts
albaranau: As for the variable name, it's in line with how Maplibre calls it (the function uses `map.cameraForBounds()` which returns CenterZoomBearing). Since it's not a Viewport with a full set info, maybe it would make sense to just call it `centerZoomBearing`?

albaranau: There are similar variables called "camera" across the code base. Maybe let's keep it like this for now, for the sake of consistency?

src/core/shared_state/currentMapPosition.ts
albaranau: Maplibre has "CameraOptions" type 
image

src/core/shared_state/currentMapPosition.ts
albaranau: Moved the function out of the atom but kept it in the same file to avoid it being used independently from it.

src/core/shared_state/currentMapPosition.ts
albaranau: Fixed

Natallia-Harshunova (approved):

src/core/shared_state/currentMapPosition.ts
Natallia-Harshunova: @albaranau hmmm, i thought that return type of signature of this function `=> mapBbox,` is the type, assigned to a state, and according to your code it also can be `CenterZoomPosition`, I just thought it can be 
 `setCurrentMapBbox: (mapBbox: Bbox | [[number, number], [number, number]]) => mapPosition,`

albaranau: @Natallia-Harshunova these functions in the first parameter are mapper functions which allow to map action parameters (i.e. modify them however we need):
https://www.reatom.dev/compat/core-v2/#basic-example

For example, I could do something like this:
image

dqunbp (approved):

src/core/shared_state/currentMapPosition.ts
Natallia-Harshunova: Now I got it, thank you for explanation

PR #902, opened 2024-11-19
bugfix(17438): fix error when there is no source when reset layer

Natallia-Harshunova:
https://kontur.fibery.io/Tasks/Task/Bivariate-renderer-errors-when-switching-on-off-MCDA-layer-17438

The reason for the error was that in the `willUnmount ` function, the call to `resetFeatureStates `(the `reset `method from `createFeatureStateHandlers`) occurred after the current source was removed from the map.
And in this `resetFeatureStates ` function there is call to the source, that has already been removed, so the function should be raised higher in the code

## Summary by CodeRabbit

- **New Features**
	- Enhanced cleanup process during unmounting by resetting feature states.
	- Improved layer styling for MCDA layers, ensuring a more consistent visual representation.

- **Bug Fixes**
	- Addressed issues with the previous unmounting logic, ensuring proper cleanup.

- **Refactor**
	- Updated method for determining layer styles to streamline the rendering process.
	- Removed unnecessary handling of map zoom events in popup methods.

### DIFF
--- a/src/core/logical_layers/renderers/BivariateRenderer/BivariateRenderer.tsx
+++ b/src/core/logical_layers/renderers/BivariateRenderer/BivariateRenderer.tsx
@@ -425,6 +425,7 @@ export class BivariateRenderer extends LogicalLayerDefaultRenderer {
     }
 
     this.cleanPopup();
+    this.resetFeatureStates?.();
 
     if (map.getSource(this._sourceId)) {
       map.removeSource(this._sourceId);
@@ -436,7 +437,6 @@ export class BivariateRenderer extends LogicalLayerDefaultRenderer {
     this.cleanUpListeners();
     map.off('zoom', this.onMapZoom);
     this.onMapZoomHandlers.clear();
-    this.resetFeatureStates?.();
   }
 
   willHide({ map }: { map: ApplicationMap }) {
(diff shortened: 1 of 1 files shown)


src/core/logical_layers/renderers/BivariateRenderer/BivariateRenderer.tsx
```diff
@@ -425,6 +425,7 @@ export class BivariateRenderer extends LogicalLayerDefaultRenderer {
     }
 
     this.cleanPopup();
+    this.resetFeatureStates?.();
```
amurKontur, line 420: image

according to the fact that at the code we have this method of 'this' added dynamically, I can be wrong, but the function might be undefined (that is why we use '?') 

a new code and also other places where this method is called do not have any handling of potential errors when function is absent 

actually what we are doing now is reseting feature states only if 'resetFeatureStates' is defined

I'd suggest doing smth like 

`this.resetFeatureStates?.() ?? console.error('No resetFeatureStates function is defined');`

or handle this error in other better ways to cover both decisions of if statement added here. If we already have one - great. Just in case we don't handle that

albaranau: @amurKontur from what I see - all the usages of `resetFeatureStates` are covered by conditional chaining (?.). 

Don't think think adding a new error message is really needed here - it's just a cleanup function, and if resetFeatureStates is not defined, there's nothing to clean up.
Plus, `resetFeatureStates()` does not return any value, so the result of condition `this.resetFeatureStates?.()` will always be false.

albaranau: As for `onMapZoomHandlers` - it looks like they're not connected to anything and do literally nothing. @Natallia-Harshunova could maybe you add a TODO comment about those (+ a ticket) or maybe just remove them altogether?

albaranau (approved):

albaranau (approved):

PR #964, opened 2025-01-22
fix(20649): Always show numbers in Bivariate hex popup

albaranau:
https://kontur.fibery.io/Tasks/Task/Show-numbers-in-bivariate-popup-by-default-20649

image

## Summary by CodeRabbit

- **Refactor**
	- Simplified rendering logic for map hex tooltips.
	- Removed feature flag dependency for displaying values in map tooltips.
	- Made `values` a required property in map hex tooltip component, ensuring consistent value display.


Natallia-Harshunova (approved):
If you 100% sure that `values ` always should be presented in such tooltip, you can also remove optional from props definition of `MapHexTooltipProps `

PR #1049, opened 2025-05-19
feat(user layers): 21025 recover create layer user flow

Natallia-Harshunova:
https://kontur.fibery.io/Tasks/Task/dn-fe-Implement-Create-user-layer-flow-MVP-21025

TO TEST THIS, `create_layer` FEATURE SHOULD BE ENABLED. You can use Oasis app for this.

Things that is not implemented in this pr and should be fix/take into consideration:
- Every time user enables/disables custom layer, /details request is triggered, it's very irritating. And features  display seems inconsistent - sometimes they show up, sometimes they don’t. It’s a bit unpredictable.
- Need to add styles for Polygon and Line, so that coloring is correct
- Code structure and architecture is really unorganized and need to be rewritten

## Summary by CodeRabbit

- **New Features**
  - Introduced a custom drawing tools widget for editing map layers, accessible via the toolbar.
  - Added new toolbar control for custom layer drawing tools.
  - Expanded API capabilities to support creating, updating, deleting layers, and saving features to layers.

- **Enhancements**
  - Improved user notifications for layer creation, updates, and editing actions.
  - Layer editing now supports immediate saving of drawn features.
  - The Edit Layer form uses a consistent "Save and draw" button label.

- **Bug Fixes**
  - Fixed form rendering to allow editing non-point geometries without displaying irrelevant input fields.

- **Refactor**
  - Simplified and clarified feature selection and editing logic.
  - Removed unused or redundant code and field types.
  - Improved naming conventions for feature selection state.

- **Style**
  - Updated side panel layout for better responsiveness.

- **Localization**
  - Added a new English translation string: "Save and draw".


src/core/toolbar/index.ts
```diff
@@ -61,6 +61,7 @@ class ToolbarImpl implements Toolbar {
           CREATE_MULTIVARIATE_CONTROL_ID,
           UPLOAD_MULTIVARIATE_CONTROL_ID,
           'EditableLayer',
+          'customLayerDrawToolsControl',
```
albaranau: Needs to be extracted into a constant

src/features/create_layer/atoms/currentEditedLayerFeatures.ts
```diff
@@ -121,6 +120,28 @@ export const currentEditedLayerFeatures = createAtom(
       });
     });
 
+    onAction('saveFeatures', (features) => {
+      schedule(async (dispatch, ctx) => {
+        if (ctx.layerId && features) {
+          try {
+            await apiClient.put<unknown>(
+              `/layers/${ctx.layerId}/items/`,
```
albaranau: Please add the endpoint to `src/core/api/layers.ts` and call from there

src/features/create_layer/atoms/currentEditedLayerFeatures.ts
```diff
@@ -121,6 +120,28 @@ export const currentEditedLayerFeatures = createAtom(
...
+              new FeatureCollection(features || null),
```
albaranau: Why do we need `|| null` here? `features` is guaranteed to be truthy because of check above:

```
if (ctx.layerId && features) {
```

Natallia-Harshunova:
@coderabbitai pause

vkozio (approved):

albaranau (approved):

PR #1052, opened 2025-05-19
Fix layer unmount race condition

Komzpa:
https://kontur.fibery.io/Tasks/Task/Selected-Overlay-is-displayed-over-Population-density-if-network-is-slow-on-startup-21226

## Summary
- ensure GenericRenderer cancels pending map updates

## Testing
- `pnpm install` *(fails: EHOSTUNREACH)*

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stability by preventing outdated updates from being applied after the renderer is unmounted or destroyed.


albaranau:
possible fix for https://kontur.fibery.io/Tasks/Task/Selected-Overlay-is-displayed-over-Population-density-if-network-is-slow-on-startup-21226
Needs to be checked and tested

Komzpa:
not a fix

PR #1071, opened 2025-05-30
feat(auth): 21648 Redirect to Map after successful login

albaranau:
https://kontur.fibery.io/Tasks/Task/Switch-page-after-login-to-map-21648

## Summary by CodeRabbit

- **New Features**
  - After login, users with the map feature enabled are redirected directly to the map page, preserving URL query parameters.

- **Refactor**
  - Improved post-login navigation to redirect users to the default route without a full page reload, maintaining URL query parameters.

### DIFF
--- a/src/core/auth/OidcSimpleClient.ts
+++ b/src/core/auth/OidcSimpleClient.ts
@@ -482,12 +482,8 @@ export class OidcSimpleClient {
     try {
       const loginOk = await this.login(user, password);
       if (loginOk) {
-        if (configRepo.get().features[AppFeature.MAP]) {
-          // load map page after succesful login
-          location.href = getAbsoluteRoute('/map') + globalThis.location.search;
-        } else {
-          location.reload();
-        }
+        // after successful login, reload the page on default route
+        location.href = getAbsoluteRoute('') + globalThis.location.search;
       }
       return true;
     } catch (e: unknown) {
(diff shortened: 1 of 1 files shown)


src/core/auth/OidcSimpleClient.ts
```diff
@@ -479,8 +480,8 @@ export class OidcSimpleClient {
     try {
       const loginOk = await this.login(user, password);
       if (loginOk) {
-        // reload to init with authenticated config and profile
-        location.reload();
+        // load map page after succesful login
+        location.href = getAbsoluteRoute('/map') + globalThis.location.search;
```
Komzpa: Need to check that map is available

src/core/auth/OidcSimpleClient.ts
```diff
@@ -480,8 +482,12 @@ export class OidcSimpleClient {
     try {
       const loginOk = await this.login(user, password);
       if (loginOk) {
-        // load map page after succesful login
-        location.href = getAbsoluteRoute('/map') + globalThis.location.search;
+        if (configRepo.get().features[AppFeature.MAP]) {
```
Komzpa: will the features be the new set of features of the logged in user, or the old set of features of the logged out user (no map on Atlas?)

vkozio (changes requested):
initial redirects should be processed in initRouter() ./src/core/router/components/Router.tsx#L72

albaranau:
> initial redirects should be processed in initRouter() ./src/core/router/components/Router.tsx#L72

After talking to Darafei and thinking a bit more, it looks like we simply need to redirect to `/` after successful login. The router will handle it and do the rest.

Applied the fix, @vkozio  could you please take a look?

src/core/auth/OidcSimpleClient.ts
albaranau: Good point. It would use the logged out feature set.
Changed the solution completely, so closing this comment.

albaranau:
@coderabbitai pause

vkozio (dismissed):

PR #1177, opened 2025-07-29
chore(mcda): 00000 add alias for mcda styles

Komzpa:
## Summary
- add `~mcda` alias for MCDA style modules
- update imports to use the new alias
- update docs and tsconfig

## Testing
- `pnpm run typecheck`
- `pnpm run test:unit` *(fails: Command failed with exit code 130 after tests passed)*

------
https://chatgpt.com/codex/tasks/task_b_6888f76b3854832688cc841d860655a6


albaranau (changes requested):
This feels arbitrary, do we really need this alias specifically?
Why the folder inside layers styles? Maybe the types are located in a bad place and we need to do some refactoring instead

